Issue #23, opened 2019-02-14
May cause cast error when use @field or @query


kiruto:
```dart
@Post(url: "api")
@FormUrlEncoded()
Future<Response<SomeResponse> postRequest({@field("param1") int param1});
```

When defined a interface using int or other types which is not a String, there may cause a cast exception in ```request.dart``` because:
```
baseRequest.bodyFields = (body as Map).cast<String, String>();
```

kiruto:
I opened a PR for fix it : https://github.com/lejard-h/chopper/pull/24

PR #24, opened 2019-02-14
Fixed: May cause cast error when use @field or @query 

kiruto:
https://github.com/lejard-h/chopper/issues/23

### DIFF
--- a/chopper_generator/lib/src/generator.dart
+++ b/chopper_generator/lib/src/generator.dart
@@ -142,7 +142,7 @@ class ChopperGenerator extends GeneratorForAnnotation<chopper.ChopperApi> {
       ];
 
       if (queries.isNotEmpty) {
-        blocks.add(_genereteMap(queries).assignFinal(_parametersVar).statement);
+        blocks.add(_generateMap(queries).assignFinal(_parametersVar).statement);
       }
 
       if (headers != null) {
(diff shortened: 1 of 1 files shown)


chopper_generator/lib/src/generator.dart
```diff
@@ -355,17 +355,17 @@ class ChopperGenerator extends GeneratorForAnnotation<chopper.ChopperApi> {
     return refer("Request").newInstance(params, namedParams);
   }
 
-  Expression _genereteMap(Map<ParameterElement, ConstantReader> queries) {
+  Expression _generateMap(Map<ParameterElement, ConstantReader> queries) {
     final map = {};
     queries.forEach((p, ConstantReader r) {
       final name = r.peek("name")?.stringValue ?? p.displayName;
-      map[literal(name)] = refer(p.displayName);
+      map[literal(name)] = refer("${p.displayName}.toString").call([]);
```
hjJunior: I think u could simplify a little, what do thing about it? 

```diff
- map[literal(name)] = refer("${p.displayName}.toString").call([]);
+ map[literal(name)] = refer("${p.displayName}").call([]);

lejard-h: I don't know what code will be generate, but something like `map['key'] = '$value'` could be cleaner

lejard-h:
Thanks for PR, can you also add a test case ?

chopper_generator/lib/src/generator.dart
kiruto: @hjJunior The type of  p.displayName is String, so ${p.displayName} is an equation of p.displayName.
But the ${p.displayName}.toString is a method, so that can generate the code like "1.toString()".

kiruto:
Ok, I'll add a test case later.

kiruto:
I created a new PR for the issue.
https://github.com/lejard-h/chopper/pull/27

PR #34, opened 2019-03-15
Converter types fix

MichaelDark:
I have just faced the issue, that if you want to convert Response from JSON string to a `List` of objects, you can not do it because converter _does not know_ about the `Body` type, and request fails on converting. 

``` 
Future<Response<Body>> _decodeResponse<Body, ToDecode>(...) async {
    ...
    final converted = await withConverter.convertResponse<ToDecode>(response);
    ...
    return converted;
  }
```

image

`Unhandled Exception: type 'Response<dynamic>' is not a subtype of type 'FutureOr<Response<List<Language>>>'`

You can play with this types, but in this case without extra generic type you cannot get working program

The main idea of solution - just to add `ResultType` to converters with renaming `ToDecode` to `ItemType`

In `Converter`:
`convertResponse<ToDecode> => convertResponse<ResultType, ItemType>`

@lejard-h If you can, polish this variant (if needed) and give a stable version 2.2.0 (fox example) as fast as possible :)

### DIFF
--- a/chopper/lib/src/base.dart
+++ b/chopper/lib/src/base.dart
@@ -88,7 +88,7 @@ class ChopperClient {
   ) async {
     if (withConverter == null) return response;
 
-    final converted = await withConverter.convertResponse<ToDecode>(response);
+    final converted = await withConverter.convertResponse<Body, ToDecode>(response);
 
     if (converted == null) {
       throw Exception("No converter found for type $ToDecode");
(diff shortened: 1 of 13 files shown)


lejard-h:
Thanks, can you fix CI ?
regenerate example, test and format the code

MichaelDark:
Sure, I'll start doing it at about 4PM UTC0 today

MichaelDark:
@lejard-h Looking forward for merge

lejard-h:
Looks good thanks :)

MichaelDark:
@lejard-h Could you publish new version to dartpub? I realy need this fix (new version v2.1.1) :)

Issue #44, opened 2019-04-16
Should ignore null values instead of sending null


humazed:
in this example

```
  @Post(path: "store/cities/search")
  Future<Response<CitiesResponse>> getCities(
    @Field("key0") int key0,
    @Field("key1") int key1,
  );

```
if I call the method with null value the value should be ignored this is the default behavior of Retrofit

I suggest the generated code should be:

```
  final $body = {
    'key0': key0,
    'key1': key1,
  }..removeWhere((key, value) => value == null);
```

if you are interested I can start working on it.

lejard-h:
Fix in release `2.3.1`

humazed:
I see the null check is only added to multipart what about the other request types? 

I see it's still sending 'null' instead of ignoring the field.

is that an oversight or deliberate design decision?

lejard-h:
Client should not transform body value
You could have server that accept null value on a field to delete it

if you don't want null value in body you should implement it at the serializer level or use a serializer that handle non nullable

humazed:
I don't think this is the default behavior of retrofit, because what's the point of sending null to the server, and in some cases, like mine, the server can't handle sending null and I have no control over it and I have no problem with it while using retrofit.

I didn't understand what you mean by 

> Client should not transform body value

in the example above I expect when calling 

` getCities("v1",null);`

the request body is: `{ "key0" : "v1" }`

not: `{ "key0" : "v1", "key1" : null }`

dongnqda:
@lejard-h @humazed Hi guys, any official way to work around this ? Currently I'm facing the same issue. Do we have any kind of annotation to not include the key if the value is null ?

Our backend is PHP Laravel, and my teammate told me that it was the way Laravel work, so I have to handle it on my side.

Currently I found 2 temp ways to work around this but it's quite odd to me.

1. Define api endpoint to receive the Map<String, dynamic> and we have to manually manipulate the Map ourself.
2. Define multiple endpoints base on input params for a single api url.

PR #45, opened 2019-04-16
master

humazed:
added fix to #44 and an example to it.


lejard-h:
Hello, thanks for PR

Can you move the logic inside `toMultipartRequest` it is simpler, and like that we don't need to rebuild

https://github.com/lejard-h/chopper/blob/master/chopper/lib/src/request.dart#L146

lejard-h:
patched in commit https://github.com/lejard-h/chopper/commit/71bc654a1829169fa8f91fbd198036cc7b86aafb

humazed:
I see the null check is only added to multipart what about the other request types? is it added in another commit or will be added?

PR #54, opened 2019-05-30
[Fixes #53] Check path length to prevent trailing slash

### LINKED ISSUE #53: Unwanted trailing slash
With the current code of the generator it is not possible to avoid having a trailing slash if for instance a get request is called on the base url of the server.

It would be nice if a trailing slash is only used when explicitly given.

Kavantix:


lejard-h:
Sorry I forgot this PR, I will try to release tonight

In an other diff can you create a test that represent your usecase ?

Kavantix:
@lejard-h Is it possible you released the wrong version of the code? 2.3.3 of the generator doesn't have the fix

lejard-h:
Sorry I had to review other PR :) the change is available on chopper (not chopper_generator) version `2.3.2`

Kavantix:
@lejard-h It is a change in the generator, but it is not in version 2.3.3

lejard-h:
My bad it should be okay now on `2.3.4`

PR #56, opened 2019-06-10
Add stream support for body request and for output body response

MatMaul:
For usage just pass a `Stream<List<int>>` as a body for requests, or specify `Response<Stream<List<int>>>` as a return type in your generator to get a stream.

### DIFF
--- a/chopper/lib/src/base.dart
+++ b/chopper/lib/src/base.dart
@@ -12,6 +12,7 @@ import "interceptor.dart";
 import "request.dart";
 import 'response.dart';
 import 'annotations.dart';
+import 'utils.dart';
 
 Type typeOf<T>() => T;
 
@@ -167,9 +168,12 @@ class ChopperClient {
     req = await _interceptRequest(req);
 
     _requestController.add(req);
-    final stream = await httpClient.send(await req.toBaseRequest());
+    final streamRes = await httpClient.send(await req.toBaseRequest());
+    if (isTypeOf<BodyType, Stream<List<int>>>()) {
+      return Response(streamRes, (streamRes.stream) as BodyType);
+    }
 
-    final response = await http.Response.fromStream(stream);
+    final response = await http.Response.fromStream(streamRes);
     Response res = Response(response, response.body);
 
     if (res.isSuccessful) {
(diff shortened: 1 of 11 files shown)


chopper/lib/src/base.dart
```diff
@@ -168,6 +169,9 @@ class ChopperClient {
 
     _requestController.add(req);
     final stream = await httpClient.send(await req.toBaseRequest());
```
lejard-h: maybe rename `final stream` to something else to avoid `stream.stream`

chopper/lib/src/interceptor.dart
```diff
@@ -115,9 +115,12 @@ class HttpLoggingInterceptor
...
 
     var bytes;
-    if (response.base.body != null && response.base.body.isNotEmpty) {
-      chopperLogger.info(response.base.body);
-      bytes = ' (${response.base.bodyBytes.length}-byte body)';
+    if (response.base is http.Response) {
+      final resp = response.base as http.Response;
+      if (resp.body != null && resp.body.isNotEmpty) {
+        chopperLogger.info(resp.body);
+        bytes = ' (${response.bodyBytes?.length}-byte body)';
```
lejard-h: `response.bodyBytes` can be null ? if so we should log something else instead of `null-byte body`

MatMaul: It can't because we test just before that `base` is a Response and not a StreamedResponse.

MatMaul:
I added one test case for each use case.

MatMaul:
Also perhaps we should put a breaking change warning in the release note because I changed the type of `base`, and `body` is not accessible through it now. People should use `bodyBytes` instead now.

lejard-h:
Indeed that can be problematic.
we could implement a `bodyString` getter to the Chopper `Response` that reimplement the `body` getter from `Response` https://pub.dev/documentation/http/latest/http/Response/body.html and yes introduce a breaking change in CHANGELOG

MatMaul:
Indeed, I'll push a change with bodyString.

MatMaul:
Done !

MatMaul:
I added the warning in the changelog and bump the release to 2.4.0.

lejard-h:
Thanks it's perfect 👍

MatMaul:
Great, and I saw you added multipart and it's already released :+1:

PR #62, opened 2019-06-25
avoid auto append '/' if baseUrl is empty 

rinlv:
for case user want to use custom url depend on path


rinlv:
Hi @lejard-h , sorry for this mistake, I have been formatted dart language and please check it.

lejard-h:
great thanks 👍

PR #78, opened 2019-08-22
Add ability to use the BodyType in response interceptor

Kavantix:
The current version of the `ResponseInterceptorFunc` did not allow me to use the `BodyType` in the actual interceptor because of how the typedef was done.
I also added the `DynamicResponseInterceptorFunc` such that old code that was using the interceptor without `BodyType` will keep working as expected

Note: it might be nice if in the future we also integrate the `BodyType` into the `onResponse` method of the `ResponseInterceptor` class

### DIFF
--- a/chopper/lib/src/base.dart
+++ b/chopper/lib/src/base.dart
@@ -20,7 +20,8 @@ final allowedInterceptorsType = <Type>[
   RequestInterceptor,
   RequestInterceptorFunc,
   ResponseInterceptor,
-  ResponseInterceptorFunc
+  ResponseInterceptorFunc,
+  DynamicResponseInterceptorFunc,
 ];
 
 /// Root object of chopper
@@ -77,7 +78,9 @@ class ChopperClient {
       value is RequestInterceptor || value is RequestInterceptorFunc;
 
   bool _isResponseInterceptor(value) =>
-      value is ResponseInterceptor || value is ResponseInterceptorFunc;
+      value is ResponseInterceptor ||
+      value is ResponseInterceptorFunc ||
+      value is DynamicResponseInterceptorFunc;
 
   bool _isAnInterceptor(value) =>
       _isResponseInterceptor(value) || _isRequestInterceptor(value);
@@ -145,6 +148,8 @@ class ChopperClient {
       if (i is ResponseInterceptor) {
         res = await i.onResponse(res);
       } else if (i is ResponseInterceptorFunc) {
+        res = await i(res);
+      } else if (i is DynamicResponseInterceptorFunc) {
         res = await i(res);
       }
     }
--- a/chopper/lib/src/interceptor.dart
+++ b/chopper/lib/src/interceptor.dart
@@ -52,8 +52,10 @@ class HeadersInterceptor implements RequestInterceptor {
       applyHeaders(request, headers);
 }
 
-typedef ResponseInterceptorFunc<Value> = FutureOr<Response> Function(
+typedef ResponseInterceptorFunc = FutureOr<Response<Value>> Function<Value>(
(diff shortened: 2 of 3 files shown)


chopper/lib/src/base.dart
```diff
@@ -144,12 +147,14 @@ class ChopperClient {
         res = await i.onResponse(res);
       } else if (i is ResponseInterceptorFunc) {
         res = await i(res);
```
lejard-h: passing the body type here should be enough
`i<BodyType>(res)`

chopper/lib/src/interceptor.dart
```diff
@@ -52,8 +52,10 @@ class HeadersInterceptor implements RequestInterceptor {
       applyHeaders(request, headers);
 }
 
-typedef FutureOr<Response> ResponseInterceptorFunc<Value>(
+typedef ResponseInterceptorFunc = FutureOr<Response> Function<Value>(
     Response<Value> response);
+typedef DynamicResponseInterceptorFunc = FutureOr<Response> Function(
```
lejard-h: didn't get why we need `DynamicResponseInterceptorFunc`, without passing a generic type `ResponseInterceptorFunc` it's already dynamic.

chopper/lib/src/interceptor.dart
```diff
@@ -52,8 +52,10 @@ class HeadersInterceptor implements RequestInterceptor {
       applyHeaders(request, headers);
 }
 
-typedef FutureOr<Response> ResponseInterceptorFunc<Value>(
+typedef ResponseInterceptorFunc = FutureOr<Response> Function<Value>(
```
lejard-h: I missed that !
We should force to not modifying the BodyType inside an interceptor by returning `FutureOr<Response<Value>>`

chopper/lib/src/interceptor.dart
Kavantix: Well the way you define the interceptor before was without a type argument, so the function definition will be strictly different from the new one.
So without the `DynamicResponseInterceptorFunc` you cannot specify an interceptor like: `FutureOr<Response> interceptor(Response r) {}` it has to be `FutureOr<Response<T>> interceptor<T>(Response<T> r) {}`

chopper/lib/src/base.dart
Kavantix: Not sure what you mean here, passing the body type is not needed

chopper/lib/src/interceptor.dart
Kavantix: Good point added it

chopper/lib/src/base.dart
lejard-h: if you don't pass it will always be dynamic, the interceptor will never not the body type of the response

chopper/lib/src/interceptor.dart
lejard-h: `typedef ResponseInterceptorFunc<Value> = FutureOr<Response> Function(Response<Value> response);`
has a type argument which is optional so you should be able to do both

Kavantix: No that definition of the interceptor does not allow you to use the body type in the actual interceptor

lejard-h: hmm I see, `dartfmt --fix` generated somthing weird :/
`typedef ResponseInterceptorFunc<Value>` instead of `typedef ResponseInterceptorFunc = Function<Value>`

chopper/lib/src/base.dart
Kavantix: The interceptor knows the type since it is inferred by dart from the response that is given (since they have the same body type value)

PR #79, opened 2019-08-22
Should assert that requests and responses are returned from interceptors

mclark4386:
also 2 typo fixes

### DIFF
--- a/chopper/lib/src/base.dart
+++ b/chopper/lib/src/base.dart
@@ -126,10 +126,15 @@ class ChopperClient {
         req = await i(req);
       }
     }
+      
+    assert(
+      req == null,
+      'Interceptors should return modified request'
+    );
 
     assert(
       body == req.body,
-      'Intercptors should not transform the body of the request'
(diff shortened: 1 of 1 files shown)


chopper/lib/src/base.dart
```diff
@@ -126,10 +126,15 @@ class ChopperClient {
         req = await i(req);
       }
     }
+      
+    assert(
+      req == null,
```
Kavantix: This comparison is wrong

chopper/lib/src/base.dart
```diff
@@ -146,10 +151,15 @@ class ChopperClient {
         res = await i(res);
       }
     }
+      
+    assert(
+      res == null,
```
Kavantix: This comparison is wrong

Kavantix:
@lejard-h can you revert this pull request (see comments) this breaks all interceptors

lejard-h:
It's already fix on master

Issue #115, opened 2020-01-13
You need a MockInterceptor() for mock data


ozburo:
How can we implement a Mock Data Interceptor like the one we have for DIO?
https://github.com/gabuldev/dio_interceptors/blob/master/lib/src/interceptor_mock.dart

I can see a way to _modify_ either the outgoing request or incoming response, but can't skip the request call (no need) and simply return a Map/Data.

Is this possible w/ Chopper? I think it would be a welcomed addition 👍

lejard-h:
Chopper use `package:http`. 

You should be able to directly use the mocking API of the HTTP package.
https://pub.dev/documentation/http/latest/testing/MockClient-class.html

ozburo:
For anyone wondering how to do this I just came back to this issue and refactored our app back to using Chopper and implemented mock data like this:

```
import 'dart:convert';

import 'package:chopper/chopper.dart';
import 'package:http/http.dart' as http;
import 'package:http/testing.dart';

part 'api_service.chopper.dart';

@ChopperApi()
abstract class ApiService extends ChopperService {
  static ApiService create() {
    final client = ChopperClient(
      client: MockClient((request) async {
        Map result = mockData[request.url.path]?.firstWhere((d) {
          if (d['id'] == request.url.queryParameters['id']) return true;
          return false;
        }, orElse: () => null);
        if (result == null) {
          return http.Response(
              json.encode({'error': "not found"}), 404);
        }
        return http.Response(json.encode(result), 200);
      }),
      baseUrl: 'https://mysite.com/api',
      services: [
        _$ApiService(),
      ],
    );
    return _$ApiService(client);
  }

  @Get(path: "/get")
  Future<Response> get(@Query() String url);
}
```

It's easy and elegant once you know how 💯

JEuler:
@ozburo Nice thing! Can we use this code for somekind of documentation about mocking? @lejard-h Are you agree? :)

ozburo:
@JEuler that's why I posted it :blush:

JEuler:
Thank you!

JEuler:
Should we close this guys? :)

JEuler:
Feel free to reopen :)

PR #133, opened 2020-04-10
Mock note

JEuler:
https://github.com/lejard-h/chopper/issues/115#issuecomment-608417333

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -44,6 +44,15 @@ Request _addQuery(Request req) {
 }
 ```
 
+## Mock ChopperClient for testing
+
+Chopper is built on top of `http` package.
+
+So, one can just use the mocking API of the HTTP package.
+https://pub.dev/documentation/http/latest/testing/MockClient-class.html
+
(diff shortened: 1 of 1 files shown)


stewemetal:
We should consider including the code sample itself in the documentation, not just the link to a comment in an issue. The comment can be referenced nontheless, but I suggest putting the code sample up front to the reader and avoiding that one indirection. :)

Also, the sample code contains this piece:

```dart
Map result = mockData[request.url.path]?.firstWhere((d) {
          if (d['id'] == request.url.queryParameters['id']) return true;
          return false;
        }, orElse: () => null);
```

That one function parameter called `d` bugs me. One-letter "names" are not names. It seems to me that something like `mockQueryParams` would be a more suitable name.

JEuler:
I am okay with oneLetter names for some cases (indexes, short code blocks) but for this code mockQueryParams is okay! :) 

I fixed!

stewemetal:
I agree, it's OK for some well-defined cases, like index variables.

Thanks for the fix! :]

Issue #122, opened 2020-02-09
token authorization


sabaya99:
how to pass bearer token authorization in chopper

vijaemanlapaz:
same here, how can we pass **Authorization** header on request via **Interceptor**

javierjmc:
Just add a request interceptor like this:

```
    ChopperClient(
       ...
        interceptors: [
          (Request request) async {
            await request.headers.putIfAbsent(
                'Authorization',
                () =>
                    'Bearer authToken');
            return request;
          }
        ],
      ));
```

`authToken` can be injected, fetched from shared preference or something like that.

vijaemanlapaz:
![Inkedcode_LI](https://user-images.githubusercontent.com/1444097/74648337-6b0bf700-51b8-11ea-8a38-5c87b8c88a6c.jpg)

above is my code
no luck, im getting the following exceptions

![Screenshot_1](https://user-images.githubusercontent.com/1444097/74648613-fe452c80-51b8-11ea-86d3-e70ce7cb9bd8.png)

javierjmc:
You need to await the Future from your sharedpreferences, otherwise by the time it answers the interception already finished.

vijaemanlapaz:
@javierjmc 
> You need to await the Future from your sharedpreferences, otherwise by the time it answers the interception already finished.

I did await for the sharedpreferences still not working.


**Update:**
Finally got it to work. here's what i ended up doing.


**ApiService.dart**
![ApiService](https://user-images.githubusercontent.com/1444097/75184805-edc81f80-577f-11ea-959b-ded6f465bff9.png)

**HeaderInterceptor.dart**
![HeaderInterceptor](https://user-images.githubusercontent.com/1444097/75184802-e86ad500-577f-11ea-81af-857d99079528.png)

JEuler:
I think this can be closed, actually I wrote a little note about this https://medium.com/@JEuler/flutter-http-authorized-requests-801e1d10debf :)

stewemetal:
I agree with @JEuler. There are countless articles about the algorithm one can implement with interceptors (mostly using Retrofit and OkHttp interceptors, but Chopper is conceptually the same) to do token authenticated requests and token refreshing.

Basically, the algorithm goes like this:

Add the authentication token to the request (by "Authentication" header, for example) -> try the request -> if it fails use the refresh token to get a new auth token -> 
1) if that succeeds, save the auth token and retry the original request with it
[...]

PR #134, opened 2020-04-11
HTTP Auth example

JEuler:
"Addressing" https://github.com/lejard-h/chopper/issues/122

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -86,4 +86,4 @@ interceptors: [
 ]
 ```
 
-Of course, the actual implementation may vary based on how the backend API of your project looks like.
+The actual implementation of the algorithm above may vary based on how the backend API - more precisely the login and session handling - of your app looks like.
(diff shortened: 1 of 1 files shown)


faq.md
```diff
@@ -61,3 +61,29 @@ final chopper = ChopperClient(
...
+    if (response?.statusCode == 401) {
+      SharedPrefs.localStorage.remove(tokenHeader);
+      // Navigate to some login page or just request new token
+    }
+    return response;
+  },
+]
+```
+
+Of course, the actual implementation may vary based on how the backend API of your project looks like.
```
stewemetal, line 89: Provided that this sentece becomes a part of the documentation, I would rephrase it like this:

The actual implementation of the algorithm above may vary based on how the backend API - more precisely the login and session handling - of your app looks like.

JEuler: Fixed, thank you!

PR #135, opened 2020-04-25
Base url note

JEuler:


faq.md
```diff
@@ -61,6 +62,22 @@ Request compressRequest(Request request) {
 Future<Response> postRequest(@Body() Map<String, String> data);
 ```
 
+## Runtime baseUrl change
+
+Sometimes you may need to change runtime base url, for example, you have server selector and you want to use different api routes like
```
stewemetal, line 67: I would rephrase this line like this:

You may need to change the base URL of your network calls during runtime, for example, if you have to use different servers or routes dynamically in your app in case of a "regular" or a "paid" user.

faq.md
```diff
@@ -61,6 +62,22 @@ Request compressRequest(Request request) {
...
+`regular` and `pro` routes. You can store the current server base url in your SharedPreferences (encrypt/decrypt it if needed) and
+use in interceptor like that:
```
stewemetal, line 69: Why is this a new line? :)

faq.md
```diff
@@ -61,6 +62,22 @@ Request compressRequest(Request request) {
...
+
+```dart
+...
+(Request request) async =>
+              SharedPreferences.containsKey('baseUrl')
+                  ? request.copyWith(
+                      baseUrl: SharedPreferences.getString('baseUrl'))
+                  : request
+...
+```
```
stewemetal, line 79: Good example, good code! 👍

faq.md
JEuler: Ooops, I'm dumb :D Just wrapped textline in VS Code :D

faq.md
JEuler: Thanks! Pushed your variant.

JEuler:
@stewemetal All is okay now? :)

JEuler:
@stewemetal ? :) Maybe @lejard-h ? :)

faq.md
```diff
@@ -61,6 +62,20 @@ Request compressRequest(Request request) {
 Future<Response> postRequest(@Body() Map<String, String> data);
 ```
 
+## Runtime baseUrl change
+
+You may need to change the base URL of your network calls during runtime, for example, if you have to use different servers or routes dynamically in your app in case of a "regular" or a "paid" user. You can store the current server base url in your SharedPreferences (encrypt/decrypt it if needed) and use in interceptor like that:
```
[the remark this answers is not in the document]
JEuler: Thanks, added your note! And squashed commits.

PR #139, opened 2020-05-08
Merge develop in master

lejard-h:
Mostly Documentations change
No need to deploy on pub


JEuler:
What about "squash and merge" setting for the PR's? Actually it is very nice :)

lejard-h:
Yes enabled squash and merge as default :)

PR #151, opened 2020-06-27
Add optionalBody parameter to suppress warnings

### LINKED ISSUE #149: Optional body for POST/PATCH/PUT
I have a `POST` method without a `Body()`, which causes the following warning:

```
[WARNING] Chopper Generator:
POST /{packageName}/edits/{editId}:commit
Body is null
Use @Body() annotation on your method parameter to provide a body to your request
   e.g.: Future<Response> postRequest(@Body() Map body);
```

I should be able to suppress this warning, which doesn't seem to be possible at the moment (looking at Chopper's source). In this case I'm calling the Google Play Developer API, which says:

> Do not supply a request body with this method.
> https://developers.google.com/android-publisher/api-ref/edits/commit

Maybe something like this is possible?

```dart
@Post(path: '/some/path', optionalBody: true)
```

LouisMT:
Add an `optionalBody` parameter to the `Post`, `Patch` and `Put` annotations to allow suppressing the warning for missing bodies in these requests.

The parameter defaults to `true` for all other HTTP methods.

Fixes #149

### DIFF
--- a/chopper/lib/src/annotations.dart
+++ b/chopper/lib/src/annotations.dart
@@ -158,8 +158,12 @@ class Method {
   /// Mark the body as optional to suppress warnings during code generation
   final bool optionalBody;
 
-  const Method(this.method,
-      {this.path = '', this.headers = const {}, this.optionalBody = true});
+  const Method(
+    this.method, {
+    this.path = '',
+    this.headers = const {},
+    this.optionalBody = true,
+  });
 }
 
 /// Defines a method as an HTTP GET request.
@@ -174,12 +178,16 @@ class Get extends Method {
 /// Use the [Body] annotation to pass data to send.
 @immutable
 class Post extends Method {
-  const Post(
-      {String path = '',
-      Map<String, String> headers = const {},
-      optionalBody = false})
-      : super(HttpMethod.Post,
-            path: path, headers: headers, optionalBody: optionalBody);
+  const Post({
+    String path = '',
+    Map<String, String> headers = const {},
+    optionalBody = false,
+  }) : super(
+          HttpMethod.Post,
+          path: path,
+          headers: headers,
+          optionalBody: optionalBody,
+        );
 }
 
 /// Defines a method as an HTTP DELETE request.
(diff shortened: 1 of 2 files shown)


JEuler:
Looks cool, guys? :)

JEuler:
But maybe set default to false?

LouisMT:
Hey, I've removed the `_methodWithBody` check. This check is redundant because of the new `optionalBody` parameter. I've also added a small hint on how to suppress the warning in the warning message itself.

> But maybe set default to false?

I've removed the default value and set it explicitly for each HTTP method. This also makes it possible for the user to override this property for all HTTP methods if they wish.

I hope this is okay for you!

chopper/lib/src/annotations.dart
```diff
@@ -155,54 +155,106 @@ class Method {
...
 @immutable
 class Delete extends Method {
-  const Delete({String path = '', Map<String, String> headers = const {}})
-      : super(HttpMethod.Delete, path: path, headers: headers);
+  const Delete({
+    optionalBody = true,
+    String path = '',
+    Map<String, String> headers = const {},
+  }) : super(HttpMethod.Delete,
+            optionalBody: optionalBody, path: path, headers: headers);
```
wilkomanger, line 209:
```suggestion
            optionalBody: optionalBody, path: path, headers: headers,);
```
Then run `dartfmt`

JEuler:
@stewemetal @lejard-h Hi guys! What are you thinking about this PR? :)

chopper/lib/src/annotations.dart
```diff
@@ -155,54 +155,110 @@ class Method {
...
+  });
 }
 
 /// Defines a method as an HTTP GET request.
 @immutable
 class Get extends Method {
-  const Get({String path = '', Map<String, String> headers = const {}})
-      : super(HttpMethod.Get, path: path, headers: headers);
+  const Get({
+    optionalBody = true,
```
lejard-h, line 173: missing type

also on Post and Delete

LouisMT:
I've specified the type for all methods now! :+1:

JEuler:
COOL! 👍

JEuler:
@stewemetal @lejard-h Hi guys! If all is cool, let's merge it? :)

lejard-h:
Nice :)

JEuler:
Cool!

PR #156, opened 2020-07-30
Fix typo and invalid code sample in documentation

BaloghTamas:
Fixing a minor typo in the docs, also I had issues with defining the interceptors the way the docs suggested. I had to define the type of the lambda parameter, also it looks like `Response` class has no `copyWith(...)` method, but has a `replace(..)` one.


stewemetal:
LGTM, thanks for the improvement!

PR #170, opened 2020-09-03
This allows chopper API methods with generic attributes

cpthooch:
e.g.:
```
  @Post(path: URL.DOCUMENT_PRIOR_INFO)
  Future<Response<T>> getPriorInfo<T>(@Body() Params params);
```

Which is usefull when single API method may respond differently
depending on request parameter values.

### DIFF
--- a/chopper_generator/lib/src/generator.dart
+++ b/chopper_generator/lib/src/generator.dart
@@ -129,6 +129,8 @@ class ChopperGenerator extends GeneratorForAnnotation<chopper.ChopperApi> {
       b.annotations.add(refer('override'));
       b.name = m.displayName;
       b.returns = Reference(m.returnType.getDisplayString());
+      b.types.addAll(m.typeParameters
+          .map((t) => Reference(t.getDisplayString(withNullability: false))));
       b.requiredParameters.addAll(m.parameters
           .where((p) => p.isNotOptional)
           .map((p) => Parameter((pb) => pb


stewemetal:
While technically the PR LGTM, I kind of doubt this should be the way. Chopper is a REST library and REST APIs' endpoints *should not* return different types of data on the same endpoint (optimally, of course) for each call to the endpoints. Also, error types (if you get an error body) should be handled in Interceptors.
What do you think guys? @lejard-h @JEuler

stewemetal:
Also, @cpthooch , what is your use-case for this?

cpthooch:
> While technically the PR LGTM, I kind of doubt this should be the way. Chopper is a REST library and REST APIs' endpoints _should not_ return different types of data on the same endpoint (optimally, of course) for each call to the endpoints. Also, error types (if you get an error body) should be handled in Interceptors.
> What do you think guys? @lejard-h @JEuler

@stewemetal I completely agree that REST APIs shouldn't behave this way, but that's not uncommon, especially in the case of legacy web services that have been developed and maintained for decades, and I'm currently on one of these. And product owners are not going to invest in the redesign at the moment.

cpthooch:
> Also, @cpthooch , what is your use-case for this?

The use case is to specify actual type of result at the caller site, e.g.:

```
final response = await getPriorInfo<SomeType>(PriorInfoParams(value: 'some value'))

final otherResponse = await getPriorInfo<SomeOtheType>(PriorInfoParams(value: 'some other value'))
```

stewemetal:
> > While technically the PR LGTM, I kind of doubt this should be the way. Chopper is a REST library and REST APIs' endpoints _should not_ return different types of data on the same endpoint (optimally, of course) for each call to the endpoints. Also, error types (if you get an error body) should be handled in Interceptors.
> > What do you think guys? @lejard-h @JEuler
> 
> @stewemetal I completely agree that REST APIs shouldn't behave this way, but that's not uncommon, especially in the case of legacy web services that have been developed and maintained for decades, and I'm currently on one of these. And product owners are not going to invest in the redesign at the moment.

OK, I can't argue with that. 😅 I'm in.

@cpthooch All right. 👍 Can you confirm that this doesn't break method generation for methods *without* type parameters?

cpthooch:
> @cpthooch All right. 👍 Can you confirm that this doesn't break method generation for methods _without_ type parameters?

Yes, I confirm! I already use this fix to generate methods with and w/o type parameters in my project.

JEuler (approved):

JEuler:
Looks cool thing to have for legacy projects! The same error with analyzer :(

JEuler:
Merging it if that's okay :)

JEuler:
@cpthooch thank you very much!

PR #178, opened 2020-10-12
Dependency update

JEuler:


JEuler:
To support the new analyzer (I didn't change anything) but tests are working (existing tests).

stewemetal:
We should do something about the Travis build failing. :/ @lejard-h Can you look into it?

JEuler:
@stewemetal I fixed, just removed the web tests, that was marked as // ignore: uri_has_not_been_generated for generated files. I think this is okay. Also, I was needed to upgrade test_coverage, now I'm pointing it to the my fork, but I have created PR on the main repo. :) https://github.com/pulyaevskiy/test-coverage/pull/24

chopper_generator/lib/src/generator.dart
```diff
@@ -129,21 +129,24 @@ class ChopperGenerator extends GeneratorForAnnotation<chopper.ChopperApi> {
     return Method((b) {
```
stewemetal, line 129: We should avoid using one-letter variable and parameter names for readability's sake. It's not the scope of this PR, so I'll let it pass, but we should really do it.

stewemetal:
LGTM! :)

stewemetal (approved):
LGTM! :)

chopper_generator/lib/src/generator.dart
JEuler: Yeah, totally agree. Let's push that after this PR. Also, we should decide how to release new versions of Chopper.

PR #181, opened 2020-10-18
Re-add Angular example and fix travis

lejard-h:
Re add example removed here (https://github.com/lejard-h/chopper/pull/178/files)

Upgrade to Angular 6, but I had to remove jaguar_serializer which is not maintain anymore (last update 1 year ago).

Then use new Dart command to run `pub`, `analyzer` and `formatter`


JEuler (approved):

JEuler:
Don't know if will be available later, so approved before Travis fix :) Thanks!

PR #189, opened 2020-11-07
The Big Public API Documentation update, volume 1.

stewemetal:
I've started to go through our public API documentation and I'm making some adjustments and doc rewrites so they can be more user friendly.

This is a work-in-progress PR and will be updated a couple of times in the following days. It will be kind of large, and I would appreciate your feedback and reviews along the way. 😃 

This PR *will not change* the public, nor the private API, it exclusively touches markdown docs and inline docs.


JEuler:
Very cooooool!

lejard-h:
Nice ! I'll take a look

chopper/lib/src/annotations.dart
```diff
@@ -262,9 +262,14 @@ class Head extends Method {
...
 }
 
+/// A function that should convert the body of a [Request] to the HTTP representation.
 typedef ConvertRequest = FutureOr<Request> Function(Request request);
+/// A function that should convert the body of a [Response] from the HTTP
+/// representation to a Dart object.
 typedef ConvertResponse<T> = FutureOr<Response> Function(Response response);
 
+/// A helper class for wrapping a request and a response converter.
+/// See [ConvertRequest], [ConvertResponse].
```
lejard-h, line 272: maybe we should add an example here

chopper/lib/src/base.dart
```diff
@@ -451,25 +473,35 @@ class ChopperClient {
...
+///     _$TodosListService(client);
+///}
+///```
 abstract class ChopperService {
   ChopperClient client;
 
-  /// Internally use to retrieve service from [ChopperClient]
-  /// FIXME: use runtimeType ?
+  /// Used internally to retrieve the service from [ChopperClient].
+  /// TODO: use runtimeType
```
lejard-h, line 501: TODO should not be part of documentation
use `//` instead of `///`

chopper/lib/src/annotations.dart
lejard-h, line 272: maybe we should have an example here

lejard-h (commented):
Awesome work !

Just added small comments

stewemetal:
> Awesome work !
> 
> Just added small comments

Thanks, I'll fix them up. :)

stewemetal:
Just to be up-to-date: I had little time to improve the docs further last week, but I'm keeping it in scope. 😉 

The public API in `chopper/lib/src` is pretty much done. If everything goes well, I'll continue with the other "modules", `chopper_built_value`, `chopper_generator`, etc.

@lejard-h, @JEuler Please do another review run on this PR. I'll open another one (volume 2.) for the next updates.

JEuler:
Of course, thank you very much! Will be able to review it later this week.
Huge work!

chopper/lib/src/base.dart
```diff
@@ -21,22 +21,25 @@ final allowedInterceptorsType = <Type>[
...
-  /// Converter call before request interceptor
-  /// and before interceptor of successful response
+  /// The [Converter] that handles request and response transformation before
+  /// the request and response interceptors are called respectively.
   final Converter converter;
 
-  /// Converter call on error request
+  /// The [ErrorConverter] that handles response transformation before the
+  /// response interceptors are called, but only on error responses
+  /// \(statusCode &lt; 200 || statusCode &gt;= 300\).
```
JEuler, line 42: I that okay? :)

JEuler (approved):
Very cool work! Thanks 👍 

As a little note - could you please create a smaller PRs for the documentation updates? It will boost up the review process. Of course, if it is convenient for you.

stewemetal:
Thanks!

I totally agree, I'll split up the next updates into smaller chunks. 😉

I'm merging this one later today and opening the next one soon.

chopper/lib/src/base.dart
stewemetal: Well, it renders as `<` and `>`, but the rendering is HTML based, so it should... 😄 But yes, it's a readability issue. I'll fix it in #193 .

stewemetal: And  `||` should be `&&` of course. 😅

Issue #47, opened 2019-05-02
Retry original request inside Interceptor


s0nerik:
OkHttp provides a [`Chain`](https://github.com/square/okhttp/wiki/Interceptors) to each interceptor. This makes it possible to easily implement failed request repetition, which is useful, for example, for refreshing access tokens.

Is there a way to replicate something like [this](https://gist.github.com/alex-shpak/da1e65f52dc916716930) with Chopper's current API?

lejard-h:
It should be possible via a Hack, but I did not test it

```dart
ResponseInterceptorFunc retryRequest = (Response response) async {
  if (response.statusCode == 401) { //if unauthorized

    // build a new http client or use the one inside chopper if available
    final client = http.Client();

    // get base request
    final request = response.base.request;

     // and retry request
    final stream = await client.send(request);
    final res = await http.Response.fromStream(stream);

    // transform http response into a chopper response
    return Response(res, res.body);
  }
  return response;
};
```

sarinupreti:
did this work ?

i am getting this error when i try to send refresh token in header .

Unhandled Exception: Connection closed before full header was received

sarinupreti:
`    (Response res) async {
      if (res.bodyString.contains("9021")) {
        // m.acquire();
        final expireTime = await prefix2.authDao.getExpireTime();
        if (DateTime.now().toUtc().millisecondsSinceEpoch > expireTime) {
          try {
            final String oldRefreshToken =
                await prefix2.authDao.getRefreshToken();

            //send old refresh token and update refresh token
            await authBloc.updateRefreshToken("Refresh " + oldRefreshToken);

            final client = http.Client();

            // get base request
            final request = res.base.request;

            // and retry request
            final stream = await client.send(request);
            final newRes = await http.Response.fromStream(stream);

            return Response(newRes, newRes.body);
          } catch (e) {
            print("Error while refreshing token" + e);
          }
        }
        // m.release();
      }

      return res;
    },`

sarinupreti:
```     (Response res) async {
      if (res.bodyString.contains("9021")) {
        // m.acquire();
        final expireTime = await prefix2.authDao.getExpireTime();
        if (DateTime.now().toUtc().millisecondsSinceEpoch > expireTime) {
          try {
            final String oldRefreshToken =
                await prefix2.authDao.getRefreshToken();

            //send old refresh token and update refresh token
            await authBloc.updateRefreshToken("Refresh " + oldRefreshToken);

[...]

PR #198, opened 2021-01-16
Implement authenticator

grahamsmith:
An attempt to solve https://github.com/lejard-h/chopper/issues/47

Inspired by OkHTTP's [authenticator](https://github.com/square/okhttp/blob/480c20e46bb1745e280e42607bbcc73b2c953d97/okhttp/src/main/kotlin/okhttp3/Authenticator.kt)

The idea is to provide a reactive authentication in the event that an auth challenge is raised. It returns a nullable Request that contains a possible update to the original Request to satisfy the authentication challenge.

e.g. a request is made to a server that responses with a 401. The authenticator can be implemented however needed to update the original request, such as using a refresh token, then pass a new request back to be executed.

### DIFF
--- a/chopper/lib/src/authenticator.dart
+++ b/chopper/lib/src/authenticator.dart
@@ -1,0 +1,9 @@
+import 'dart:async';
+
+import 'package:chopper/chopper.dart';
+
+/// Returns a request that includes a credential to satisfy an authentication challenge in
+/// [response]. Returns null if the challenge cannot be satisfied.
+abstract class Authenticator {
+  FutureOr<Request> authenticate(Request request, Response response);
+}
--- a/chopper/lib/src/authenticator.dart
+++ b/chopper/lib/src/authenticator.dart
@@ -2,8 +2,8 @@ import 'dart:async';
 
 import 'package:chopper/chopper.dart';
 
-/// Returns a request that includes a credential to satisfy an authentication challenge in
-/// [response]. Returns null if the challenge cannot be satisfied.
+/// This method should return a [Request] that includes credentials to satisfy an authentication challenge received in
+/// [response]. It should return `null` if the challenge cannot be satisfied.
 abstract class Authenticator {
   FutureOr<Request> authenticate(Request request, Response response);
 }
--- a/chopper/lib/src/base.dart
+++ b/chopper/lib/src/base.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'package:chopper/src/authenticator.dart';
 import 'package:meta/meta.dart';
 import 'package:http/http.dart' as http;
 import 'constants.dart';
@@ -36,6 +37,10 @@ class ChopperClient {
   /// The [Converter] that handles request and response transformation before
   /// the request and response interceptors are called respectively.
   final Converter converter;
+
+  /// The [Authenticator] that handles provides reactive authentication for a
+  /// request.
+  final Authenticator authenticator;
 
   /// The [ErrorConverter] that handles response transformation before the
   /// response interceptors are called, but only on error responses
@@ -109,6 +114,7 @@ class ChopperClient {
     this.baseUrl = '',
     http.Client client,
     Iterable interceptors = const [],
+    this.authenticator,
     this.converter,
     this.errorConverter,
     Iterable<ChopperService> services = const [],
@@ -315,6 +321,15 @@ class ChopperClient {
 
     final response = await http.Response.fromStream(streamRes);
     dynamic res = Response(response, response.body);
+
+    if(authenticator != null) {
+
+      var updatedRequest = authenticator.authenticate(request, res);
+
+      if(updatedRequest != null) {
+        res = await send(updatedRequest);
+      }
+    }
 
     if (_responseIsSuccessful(response.statusCode)) {
       res = await _handleSuccessResponse<BodyType, InnerType>(
--- a/chopper/lib/src/base.dart
+++ b/chopper/lib/src/base.dart
@@ -38,7 +38,7 @@ class ChopperClient {
   /// the request and response interceptors are called respectively.
   final Converter converter;
 
-  /// The [Authenticator] that handles provides reactive authentication for a
+  /// The [Authenticator] that can provide reactive authentication for a
   /// request.
   final Authenticator authenticator;
 
@@ -322,11 +322,10 @@ class ChopperClient {
     final response = await http.Response.fromStream(streamRes);
     dynamic res = Response(response, response.body);
 
-    if(authenticator != null) {
-
+    if (authenticator != null) {
       var updatedRequest = authenticator.authenticate(request, res);
 
-      if(updatedRequest != null) {
+      if (updatedRequest != null) {
         res = await send(updatedRequest);
       }
     }


grahamsmith:
I can add some tests (with some help) if the general idea is accepted.

JEuler:
Hi! Cool! Could you show please the example of usage? So, for example, I have refresh_token and refreshToken endpoint, how I should implement the Authenticator?

lejard-h:
Not sure to understand why this would allow retrying request, maybe the name `Authenticator` is confusing. I am not familiar with okHTTP, do you have an example ?

grahamsmith:
Basic example:

```
class MyAuthenticator extends Authenticator {
  
  @override
  FutureOr<Request> authenticate(Request request, Response<dynamic> response) async{
    
    if(response.statusCode == 401) {
      
      var newToken = await refreshToken();
      
      var updatedHeaders = Map.of(request.headers);

      updatedHeaders.update('Authorization', (value) => newToken, ifAbsent: () => newToken);
      
      return request.copyWith(headers: updatedHeaders);
    } else {
      return null;
    }
  }
  
  Future<String> refreshToken() async{
    //refresh the accessToken using refreshToken however needed....
    //this could be done either via http client or a ChopperService or a repository could be a dependency, this approach is intentionally not opinionated about how this works.
  }
}
```

There are strategies around handling infinite loops. Some include via headers, checking paths or counting retries and many more options. This implementation is designed to be un-opinionated as everyone's implementations will have their own nuances.

I think the name makes sense.

Plenty of tutorials for OkHttp that are relatable. I would imagine a lot of mobile devs moving to Flutter will reference it due to its popularity (over 39,000 stars on Github).

https://github.com/square/okhttp/blob/480c20e46bb1745e280e42607bbcc73b2c953d97/okhttp/src/main/kotlin/okhttp3/Authenticator.kt
https://medium.com/@olempico/okhttp-authenticator-lets-be-secure-in-an-elegant-way-3bf6096752dc
https://objectpartners.com/2018/06/08/okhttp-authenticator-selectively-reauthorizing-requests/
https://blog.coinbase.com/okhttp-oauth-token-refreshes-b598f55dd3b2
https://www.lordcodes.com/articles/authorization-of-web-requests-for-okhttp-and-retrofit
https://stackoverflow.com/questions/22490057/android-okhttp-with-basic-authentication
http://sangsoonam.github.io/2019/03/06/okhttp-how-to-refresh-access-token-efficiently.html
https://medium.com/tiendeo-tech/android-refreshing-token-proactively-with-okhttp-interceptors-e7b90c47f5e7
https://medium.com/@sandeeptengale/problem-solved-2-access-token-refresh-with-okhttp-authenticator-5ccb798ede70

grahamsmith:
I think there is also some work to do around not invoking the interceptors again. I wanted to start simple before getting into this and see if there was buy in.

stewemetal:
This is a good approach IMHO. Users of Chopper can utilize the `Request`/`Response` API to implement retry with authentication logic without depending on http directly.

It seems to be in line with OkHttp's solution linked above.

I'll do a review on the current implementation.

If this is merged, documentation should be updated with explanation on how the Authenticator should be used with example(s).

grahamsmith:
@stewemetal - Is there anything I can do to help push this along? I really could do with this functionality in my own project. Noticed it was added to the milestone which is cool :+1:

JEuler:
I think this can be approved, as I saw your example. "Infinite" retries is a local problem, that can be solved per project. Looking cool. 

But, @stewemetal, please also have a look! :)

JEuler (approved):

chopper/lib/src/authenticator.dart
```diff
@@ -0,0 +1,9 @@
+import 'dart:async';
+
+import 'package:chopper/chopper.dart';
+
+/// Returns a request that includes a credential to satisfy an authentication challenge in
+/// [response]. Returns null if the challenge cannot be satisfied.
```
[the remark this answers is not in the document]
stewemetal: So the class needs a proper, more generic description too. I'd even go as far as putting a link to OkHttp's similar mechanic (like you did in the PR).

chopper/lib/src/base.dart
```diff
@@ -37,6 +38,10 @@ class ChopperClient {
   /// the request and response interceptors are called respectively.
   final Converter converter;
 
+  /// The [Authenticator] that handles provides reactive authentication for a
```
stewemetal, line 41: "handles provides" -> "can provide"

chopper/lib/src/base.dart
```diff
@@ -316,6 +322,15 @@ class ChopperClient {
     final response = await http.Response.fromStream(streamRes);
     dynamic res = Response(response, response.body);
 
+    if(authenticator != null) {
+
```
stewemetal, line 326: This empty line is unnecessary.

stewemetal (changes requested):

stewemetal:
I've don a quick review on the PR. 

@grahamsmith Please fix the mentioned things. 😉

stewemetal:
Also, I'll find a place for the example implementation in the updated docs, it looks good. 😄

stewemetal:
> @stewemetal - Is there anything I can do to help push this along? I really could do with this functionality in my own project. Noticed it was added to the milestone which is cool 👍

I'd like to consult with @lejard-h and @JEuler soon on the topic of release management and releasing 3.0.7 with revamped docs and guides. As the Authenticator mechanic is a non-breaking change, it can indeed be part of 3.0.7 (hence the inclusion to the milestone).

JEuler:
As I said, I feel like it is a cool, non-breaking, additional mechanic, so, no problem with adding it to the upcoming release 👍

grahamsmith:
@stewemetal happy to make the changes. For future PRs I think it would be quicker to just make the changes (I believe maintainers are allowed to contribute back to my fork) rather than me copy paste them in.

JEuler:
@grahamsmith I've added changes requested from @stewemetal :) Without the description.

JEuler (approved):

grahamsmith:
Amazing thanks! I've been having a tough week so that's very much appreciated 🎉😄

JEuler:
@stewemetal Do you want us to make this change with the description? Or we are okay to merge it as it is? :)

stewemetal:
I think it's OK, I'll approve it. 😄 
I'll add the extra description in another docs update PR.

stewemetal (approved):

Aqluse:
image
You guys sure that this check won't let us fall to a endless recursion? Because I seem to get the one

grahamsmith:
@Aqluse yes someone has broken the implementation I wrote.

It was originally

```
var updatedRequest = authenticator.authenticate(request, res);

      if(updatedRequest != null) {
        res = await send(updatedRequest);
      }

```

@stewemetal and friends this will now result in an infinite loop with current implementation.

JEuler:
Let's fix it then :)

grahamsmith:
I'm on holiday at the moment so will leave to those with Dev laptops and reliable internet.

stewemetal:
Thanks for pointing this out! This implementation has gone awkward at some point during the null safety migration.

Would this change fix your use-case? It breaks the current Authenticator implementations, but it's just a first try on a quick fix:

```dart
abstract class Authenticator {
  FutureOr<Request?> authenticate(Request request, Response response);
}


// send()
...
if (authenticator != null) {
  var updatedRequest = await authenticator!.authenticate(request, res);

  if (updatedRequest != null) {
    res = await send(updatedRequest);
  }
}
...

grahamsmith:
Looks good

stewemetal:
Fixed in #241

PR #201, opened 2021-02-16
Feature/support body get

alfredo-handcash:
Hi! Thanks for your time creating this awesome HTTP client! 🍾

I added the possibility of pass body in GET request that are not implemented yet.

### DIFF
--- a/chopper/lib/src/base.dart
+++ b/chopper/lib/src/base.dart
@@ -349,6 +349,7 @@ class ChopperClient {
   /// Makes a HTTP GET request using the [send] function.
   Future<Response<BodyType>> get<BodyType, InnerType>(
     String url, {
+    dynamic body,
     Map<String, String> headers,
     Map<String, dynamic> parameters,
     String baseUrl,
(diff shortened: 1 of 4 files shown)


JEuler:
Hi! Sorry for the delay, but as I understand, this is a bad practice, to have BODY in GET. @stewemetal What do you think? Should we provide that? :)

JEuler:
@stewemetal ? :)

stewemetal:
Sorry, I thought I answered this, but I forgot to send the answer as it seems... Got my hands a little full nowadays. 😅 

So:
Unfortunately, I saw quite a few backend solutions that require the clients to send objects in GET bodies... I strongly believe that this **is** antipattern. However, supporting people who have to deal with such APIs won't hurt us. 🙂 Maybe we could place a small comment on the property (and in the documentation of GET) that this in fact is supported, but highly discouraged.

JEuler (approved):

JEuler:
Okay, thank you! Approving then 👍

PR #214, opened 2021-03-11
feature/updated doc

dafinoer:
hi, im added some setup for use the proxy. what do you think?

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -133,6 +133,7 @@ import 'dart:io';
 import 'package:http/io_client.dart' as http;
 
 final ioc = new HttpClient();
+httpx.findProxy = (url) => 'PROXY 192.168.0.102:9090';
 ioc.badCertificateCallback = (X509Certificate cert, String host, int port)
   => true;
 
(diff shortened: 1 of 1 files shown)


faq.md
```diff
@@ -133,6 +133,7 @@ import 'dart:io';
 import 'package:http/io_client.dart' as http;
 
 final ioc = new HttpClient();
+httpx.findProxy = (url) => 'PROXY 192.168.0.102:9090';
```
stewemetal, line 136: There's an extra 'x' here. Remove it and the PR will be good to go. 😉

stewemetal (changes requested):

faq.md
dafinoer: upss, im sorry, 😆

stewemetal: No harm done, thanks for the contribution! 😉 I also made a mistake telling you that it should be `http`, but in fact it should have been `ioc` as you fixed it. 😅

stewemetal (approved):

PR #241, opened 2021-04-09
Fix infinite loop when using Authenticators

stewemetal:
As pointed out in the comments of #198, the Authenticator implementation went wrong along the way to null safety migration.

This PR attempts to fix that.

### DIFF
--- a/chopper/lib/src/authenticator.dart
+++ b/chopper/lib/src/authenticator.dart
@@ -5,5 +5,5 @@ import 'package:chopper/chopper.dart';
 /// This method should return a [Request] that includes credentials to satisfy an authentication challenge received in
 /// [response]. It should return `null` if the challenge cannot be satisfied.
 abstract class Authenticator {
-  FutureOr<Request> authenticate(Request request, Response response);
+  FutureOr<Request?> authenticate(Request request, Response response);
 }
--- a/chopper/lib/src/base.dart
+++ b/chopper/lib/src/base.dart
@@ -311,9 +311,11 @@ class ChopperClient {
     dynamic res = Response(response, response.body);
 
     if (authenticator != null) {
-      var updatedRequest = authenticator!.authenticate(request, res);
+      var updatedRequest = await authenticator!.authenticate(request, res);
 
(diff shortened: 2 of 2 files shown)


JEuler (approved):

Taldorr:
Hi,
I recently tried out chopper and highly appreciate the work you guys do here.
When implementing an authenticator, I got into an infinite loop and found this issue closed.
From my understanding, as there is an authenticator involved, `updatedRequest` will never be null so the method will call itself recursively.

 What am I doing wrong?

chopper 6.6.1
chopper_generator 6.0.0

Method "send" from chopper's base.dart

`Future<Response<BodyType>> send<BodyType, InnerType>(
    Request request, {
    ConvertRequest? requestConverter,
    ConvertResponse? responseConverter,
  }) async {
    final Request req = await _interceptRequest(
      await _handleRequestConverter(request, requestConverter),
    );

    final Request req = await _interceptRequest(
      await _handleRequestConverter(request, requestConverter),
    );
    _requestController.add(req);

    final streamRes = await httpClient.send(await req.toBaseRequest());
    if (isTypeOf<BodyType, Stream<List<int>>>()) {
      return Response(streamRes, (streamRes.stream) as BodyType);
    }

    dynamic res = Response(response, response.body);

    if (authenticator != null) {
      final Request? updatedRequest = await authenticator!.authenticate(
        request,
        res,
        request,
      );

      if (updatedRequest != null) {
        res = await send<BodyType, InnerType>(
          updatedRequest,
          requestConverter: requestConverter,
          responseConverter: responseConverter,
        );
        // To prevent double call with typed response
        if (_responseIsSuccessful(res.statusCode)) {
          return _processResponse(res);
        } else {
          res = await _handleErrorResponse<BodyType, InnerType>(res);

          return _processResponse(res);
        }
      }
    }

    res = _responseIsSuccessful(res.statusCode)
        ? await _handleSuccessResponse<BodyType, InnerType>(
            res,
            responseConverter,
          )
        : await _handleErrorResponse<BodyType, InnerType>(res);

    return _processResponse(res);
  }
`

Issue #259, opened 2021-05-19
Recursive call of send for Authenticator updated request is type erased


imatok:
Hello, 

when using Authenticator to refresh access token and update the original unauthorised failing request, I have an issue with my custom Response Converter cause response BodyType and InnerType are not propagated in that recursive send. 
```dart
      if (updatedRequest != null) {
        res = await send(updatedRequest);
      }
```
Problem is, that recursive call succeeds and this `_handleSuccessResponse<BodyType, InnerType>` is executed without types:
```dart
    if (_responseIsSuccessful(response.statusCode)) {
      res = await _handleSuccessResponse<BodyType, InnerType>(
        res,
        responseConverter,
      );
    } else {
      res = await _handleErrorResponse<BodyType, InnerType>(res);
    }
```
that fails in my Response Converter that cannot resolve to correct BodyType/InnerType.
Also, after the recursive call is successfully done, returned response is not tested by this first line above: `if (_responseIsSuccessful(response.statusCode))` but the old final one: 
```dart
    final response = await http.Response.fromStream(streamRes);
``` 
so, the execution of original request proceeds with that `else` error path. 

Quick fix would be to change line 317 to:
```dart
   res = await send<BodyType, InnerType>(updatedRequest);
````
and line 321 to:
```dart
    if (_responseIsSuccessful(res.statusCode)) {
```
but I haven't run any of your tests, except trying this out in my own code. :)
Additionally, when using this recursive call, same response gets processed twice, 1st time in recursive call and then when recursion returns replaced `res` value.

BR.

JEuler:
https://github.com/lejard-h/chopper/pull/263

JEuler:
Released fix in 4.0.1

PR #263, opened 2021-05-30
Fix for #259

JEuler:
Credits to @imatok. (https://github.com/lejard-h/chopper/issues/259). Tests are passing and the code seems valid.


lejard-h (approved):
:+1:

PR #276, opened 2021-07-22
Prevent double call on token refreshment

JEuler:
There is double processing when you've got your authenticator call. It leads to exceptions like "BuiltList<YourType> is not a String" or something. 

https://github.com/lejard-h/chopper/issues/259 contains the same behavior description.

### DIFF
--- a/chopper/lib/src/base.dart
+++ b/chopper/lib/src/base.dart
@@ -315,6 +315,13 @@ class ChopperClient {
 
       if (updatedRequest != null) {
         res = await send<BodyType, InnerType>(updatedRequest);
+        // To prevent double call with typed response
+        if (_responseIsSuccessful(res.statusCode)) {
+          return _processResponse(res);
+        } else {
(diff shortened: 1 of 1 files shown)


chopper/lib/src/base.dart
```diff
@@ -327,10 +334,14 @@ class ChopperClient {
       res = await _handleErrorResponse<BodyType, InnerType>(res);
     }
 
-    res = await _interceptResponse<BodyType, InnerType>(res);
+    return _processResponse(res);
+  }
 
+  /// Utitlity method to shrink duplicated code (interceptors processing)
```
stewemetal, line 340: I think this comment is unnecessary. 😉

stewemetal (changes requested):

chopper/lib/src/base.dart
JEuler: Fixed!

stlr00:
When it will be at pub.dev?

PR #320, opened 2022-01-10
Analyzer upgrade

JEuler:


stewemetal:
The PR itself is OK of course, but the analyzer build step is failing 😕

JEuler:
Yeah, we need to rework actions. Let's go in this PR.

stewemetal:
@JEuler CI is green, we're ready to deploy 💪

stewemetal (approved):

JEuler:
Awesome job! Thank you! One question:
expect(converted is Response<String>, isTrue); - why such checks are removed? :)

stewemetal:
> Awesome job! Thank you! One question: expect(converted is Response, isTrue); - why such checks are removed? :)

Because - as the Analyzer correctly pointed it out -, since `converted` is always a `Response`, the check does not provide any added value to the test. 

`jsonConverter.convertResponse<K, SV>(res)` is type safe, at least in a way that it always returns a `Response`.

JEuler:
Cooool!

PR #331, opened 2022-02-22
migrate example to nullsafety

ipcjs:
angular is not compatible with analyzer 3.0.0, so remove it.


JEuler:
@stewemetal What do you think? I think that angular example is not important

stewemetal:
@JEuler 
Well, AngularDart is in maintenance-only mode since May 2021 according to it's readme: https://github.com/angulardart

I think we can get rid of the Angular example. 😄

JEuler (approved):

JEuler:
Thank you very much!

Issue #343, opened 2022-07-08
Nullable QueryMap fails to compile


techouse:
Hi, I really love your package, however, I came across an issue where I wanted to pass a `QueryMap` as an optional/nullable value and this resulted in it simply being added to the list of `$param` using `$params.addAll(theNullableQueryMap)`. This however fails, because a nullable value can not be added to a `Map` using [`Map.addAll`](https://api.flutter.dev/flutter/dart-core/Map/addAll.html).

Essentially this:
```dart
@Get(path: 'articles')
Future<Response<List<Article>>> getArticles({
  @Query('article_id') String? articleId,
  @Query('author_name') String? authorName,
  @Query('limit') int? limit = 10,
  @QueryMap() Map<String, dynamic>? filters,
});
```

results in
```dart
@override
Future<Response<List<Article>>> getArticles(
    {String? articleId,
    String? authorName,
    int? limit = 10,
    Map<String, dynamic>? filters}) {
  final $url = 'articles';
  final $params = <String, dynamic>{
    'article_id': articleId,
    'authorName': authorName,
    'limit': limit
  };
  $params.addAll(filters); // <-- this here will fail to compile because `filters` can't be nullable 
  final $request = Request('GET', $url, client.baseUrl, parameters: $params);
  return client.send<List<Article>, Article>($request);
}
```

It throws this compilation error:
```
Error: The argument type 'Map<String, dynamic>?' can't be assigned to the parameter type 'Map<String, dynamic>' because 'Map<String, dynamic>?' is nullable and 'Map<String, dynamic>' isn't.

- 'Map' is from 'dart:core'.
   $params.addAll(filters);
                  ^
```

I would propose it checks if `filters` is `null` before using `$params.addAll` 

```dart
final $params = <String, dynamic>{
  'article_id': articleId,
  'authorName': authorName,
  'limit': limit
};
if (filters != null) {
  $params.addAll(filters);
}
```

or simply using something more elegant like:
```dart
final $params = <String, dynamic>{
  'article_id': articleId,
  'authorName': authorName,
  'limit': limit,
  ...?filters,
};
```

PR #344, opened 2022-07-10
Fix nullable QueryMap fails to compile

techouse:
This addresses https://github.com/lejard-h/chopper/issues/343

I've used [Expression.ifNullThen](https://pub.dev/documentation/code_builder/latest/code_builder/Expression/ifNullThen.html) to additionally check for a nullable `QueryMap`.

This should be able to take

```dart
@Get(path: 'query_map')
Future<Response> getQueryMapTest4({
  @Query('name') String name = '',
  @Query('number') int? number,
  @QueryMap() Map<String, dynamic>? filters, // <-- nullable / optional filters
});
```

and generate

```dart
@override
Future<Response<dynamic>> getQueryMapTest4(
    {String name = '', int? number, Map<String, dynamic>? filters}) {
  final $url = '/test/query_map';
  final $params = <String, dynamic>{'name': name, 'number': number};
  $params.addAll(filters ?? {}); // <-- because Map.addAll expects a non-nullable Map
  final $request = Request('GET', $url, client.baseUrl, parameters: $params);
  return client.send<dynamic, dynamic>($request);
}
```

without any subsequent compilation errors.

Before this hotfix one had to use a default `const {}` like this in order to omit the `QueryMap` and avoid a compilation error

```dart
@Get(path: 'query_map')
Future<Response> getQueryMapTest3({
  @Query('name') String name = '',
  @Query('number') int? number,
  @QueryMap() Map<String, dynamic> filters = const {}, // <-- default
});
```

which would generate

```dart
@override
Future<Response<dynamic>> getQueryMapTest3(
    {String name = '',
    int? number,
    Map<String, dynamic> filters = const {}}) { // <-- default
  final $url = '/test/query_map';
  final $params = <String, dynamic>{'name': name, 'number': number};
  $params.addAll(filters);
  final $request = Request('GET', $url, client.baseUrl, parameters: $params);
  return client.send<dynamic, dynamic>($request);
}
```

~~However, this fix will also add a null check `??` when the value is not nullable and that's why I've added the extra ignore flag `dead_null_aware_expression` in order to silence the linter in a case like this~~

```dart
@Get(path: 'query_map')
Future<Response> getQueryMapTest2(
  @QueryMap() Map<String, dynamic> query, {
  @Query('test') bool? test,
});
```

~~which generates~~

```dart
@override
Future<Response<dynamic>> getQueryMapTest2(Map<String, dynamic> query,
    {bool? test}) {
  final $url = '/test/query_map';
  final $params = <String, dynamic>{'test': test};
  $params.addAll(query ?? {}); // <-- dead_null_aware_expression because `query` is not nullable
  final $request = Request('GET', $url, client.baseUrl, parameters: $params);
  return client.send<dynamic, dynamic>($request);
}
```

~~As such, this should be regarded more or less as a hotfix as some more robust null checking should be performed on the `QueryMap` itself. At this point, I lack the expertise to do this with `source_gen` and would appreciate any further guidance 😊~~

EDIT: This issue has now been rectified. See https://github.com/lejard-h/chopper/pull/344#issuecomment-1180750751

Oh, and of course, I've added a few test cases. ☑️

### DIFF
--- a/chopper/example/definition.chopper.dart
+++ b/chopper/example/definition.chopper.dart
@@ -6,7 +6,7 @@ part of 'definition.dart';
 // ChopperGenerator
 // **************************************************************************
 
-// ignore_for_file: always_put_control_body_on_new_line, always_specify_types, dead_null_aware_expression, prefer_const_declarations, unnecessary_brace_in_string_interps
+// ignore_for_file: always_put_control_body_on_new_line, always_specify_types, prefer_const_declarations, unnecessary_brace_in_string_interps
 class _$MyService extends MyService {
   _$MyService([ChopperClient? client]) {
     if (client == null) return;
--- a/chopper/example/definition.chopper.dart
+++ b/chopper/example/definition.chopper.dart
@@ -6,7 +6,7 @@ part of 'definition.dart';
 // ChopperGenerator
 // **************************************************************************
 
-// ignore_for_file: always_put_control_body_on_new_line, always_specify_types, prefer_const_declarations
+// ignore_for_file: always_put_control_body_on_new_line, always_specify_types, dead_null_aware_expression, prefer_const_declarations, unnecessary_brace_in_string_interps
 class _$MyService extends MyService {
   _$MyService([ChopperClient? client]) {
     if (client == null) return;
@@ -18,7 +18,7 @@ class _$MyService extends MyService {
 
   @override
   Future<Response<dynamic>> getResource(String id) {
-    final $url = '/resources/$id';
+    final $url = '/resources/${id}';
     final $request = Request('GET', $url, client.baseUrl);
     return client.send<dynamic, dynamic>($request);
   }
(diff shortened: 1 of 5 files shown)


JEuler:
Thank you sooooooo much!

JEuler (approved):

JEuler:
Please reformat code :)

techouse:
> Please reformat code :)

Done. My bad :)

techouse:
I just found out how to do a more precise null check on the parameters. I basically built a list of all the positional optional and named parameters like

```dart
// Build an iterable of all the parameters that are nullable
final Iterable<String> optionalNullableParameters = [
  ...m.parameters.where((p) => p.isOptionalPositional),
  ...m.parameters.where((p) => p.isNamed),
].where((el) => el.type.isNullable).map((el) => el.name);
```

And then I can simply check
```dart
if (optionalNullableParameters.contains(queryMap.keys.first)) {
  // do nullable stuff
} else {
  // do non nullable stuff
}
```

This way I can remove `dead_null_aware_expression` as this update to the fix will prevent it from happening.

Issue #158, opened 2020-08-07
Utilise `compute()`


nodinosaur:
With Chopper is there a way that we can utilise `Isolates` / `compute()` for loading and parsing large documents?

https://flutter.dev/docs/cookbook/networking/background-parsing

https://api.flutter.dev/flutter/dart-isolate/Isolate-class.html

I have seen that some Devs are moving to [Dio](https://pub.dev/packages/dio) as it can already do this 👉 [README.md](https://github.com/flutterchina/dio/blob/master/README.md#in-flutter), but I am quite invested in Chopper and don't really want to rework my network layer.

Is it possible, or could such a feature be added? 

Thanks!

nodinosaur:
I forgot to mention:
https://github.com/flutterchina/dio_flutter_transformer

DevNico:
Did anything ever get merged / what is the way to go if usage of compute is desired?

nodinosaur:
I am still hoping that there will be some official documentation with a recommended example on this, I could not get the above example to work 🤔

JanKn:
I got it working, I can see isolates being spawned and the data is loaded in our app.
Let me know if that works for you too @nodinosaur .

Two steps:

1. Creating ComputeConverter:

```
class ComputeConverter extends Converter {
  ComputeConverter(this.converter);

  final JsonSerializableConverter converter;

  @override
  FutureOr<Request> convertRequest(Request request) {
    return converter.convertRequest(request);
  }

  @override
  FutureOr<Response<BodyType>> convertResponse<BodyType, InnerType>(Response response) {
    return compute((Response response) => converter.convertResponse<BodyType, InnerType>(response), response);
  }
}
```

2. And then wrapping the JsonSerializableConverter with ComputeConverter in the Chopper Client setup:

```
final client = ChopperClient(
      baseUrl: baseUrl,
      converter: ComputeConverter(const JsonSerializableConverter(
        {
          XData: XData.fromJson,
          ...
        },
      )),
      interceptors: [
        HttpRequestLoggingInterceptor(),
      ],
    );
```

vaetas:
Hey everybody,
I want to note one thing: `compute` starts a new Isolate for each call. Starting new Isolate may take between 100-200 milliseconds. Each Isolate also copies some memory therefore having many Isolates might increase memory load of your app.

[...]

PR #345, opened 2022-07-12
Change return type of decodeJson to FutureOr in order to be able to support compute()

techouse:
This partially addresses https://github.com/lejard-h/chopper/issues/158 and is mostly based on @lejard-h's suggestion https://github.com/lejard-h/chopper/issues/158#issuecomment-678760392

**This is a breaking change** as it changes the response types of some `JsonConverter` methods, namely:

- `decodeJson` now returns `FutureOr<Response>`
  ```dart
  /// before
  Response decodeJson<BodyType, InnerType>(Response response) {
    /* sync decode stuf */
  }

  /// after
  FutureOr<Response> decodeJson<BodyType, InnerType>(Response response) async {
    /* async decode stuff */
  }
  ```

- `convertResponse` now returns `FutureOr<Response<BodyType>>`
  ```dart
  /// before
  @override
  Response<BodyType> convertResponse<BodyType, InnerType>(Response response) {
    /* sync convert stuff */
  }

  /// after
  @override
  FutureOr<Response<BodyType>> convertResponse<BodyType, InnerType>(Response response) async {
    /* async convert stuff */
  }
  ```

- `tryDecodeJson` now returns `FutureOr<dynamic>` and is now `public`
  ```dart
  /// before
  dynamic _tryDecodeJson(String data) {
    /* decode json */
  }

  /// after
  FutureOr<dynamic> tryDecodeJson(String data) {
    /* now you can override this method and use compute to decode json */
  }
  ```

- `convertError` now returns `FutureOr<Response>`
  ```dart
  /// before
  @override
  Response convertError<BodyType, InnerType>(Response response) {
    /* sync decode error */
  }

  /// after
  @override
  FutureOr<Response> convertError<BodyType, InnerType>(Response response) async {
    /* async decode error */
  }
  ```

- `responseFactory` now returns `FutureOr<Response<BodyType>>`
  ```dart
  /// before
  static Response<BodyType> responseFactory<BodyType, InnerType>(Response response) {
    /* stuff */
  }

  /// after
  static FutureOr<Response<BodyType>> responseFactory<BodyType, InnerType>(Response response){
    /* stuff */
  }
  ```

### DIFF
--- a/chopper/lib/src/interceptor.dart
+++ b/chopper/lib/src/interceptor.dart
@@ -279,6 +279,7 @@ class JsonConverter implements Converter, ErrorConverter {
           Response response) async =>
       (await decodeJson<BodyType, InnerType>(response)) as Response<BodyType>;
 
+  @protected
   FutureOr<dynamic> tryDecodeJson(String data) {
     try {
       return json.decode(data);
--- a/chopper/lib/src/interceptor.dart
+++ b/chopper/lib/src/interceptor.dart
@@ -247,7 +247,7 @@ class JsonConverter implements Converter, ErrorConverter {
     return request;
   }
 
-  Response decodeJson<BodyType, InnerType>(Response response) {
+  FutureOr<Response> decodeJson<BodyType, InnerType>(Response response) async {
     final supportedContentTypes = [jsonHeaders, jsonApiHeaders];
 
     final contentType = response.headers[contentTypeKey];
@@ -264,7 +264,7 @@ class JsonConverter implements Converter, ErrorConverter {
       body = utf8.decode(response.bodyBytes);
     }
 
-    body = _tryDecodeJson(body);
+    body = await tryDecodeJson(body);
     if (isTypeOf<BodyType, Iterable<InnerType>>()) {
       body = body.cast<InnerType>();
     } else if (isTypeOf<BodyType, Map<String, InnerType>>()) {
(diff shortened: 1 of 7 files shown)


techouse:
This PR most likely needs further scrutiny as it is breaking.

vaetas:
Is there a reason to use `FutureOr` instead of `Future` in this context? I somehow understand the difference between the two. I am not sure about the technical implications here. I am very curious though.

The API request itself must be `Future`. Does it therefore matter that inner methods are `FutureOr`?

techouse:
> Is there a reason to use `FutureOr` instead of `Future` in this context? I somehow understand the difference between the two. I am not sure about the technical implications here. I am very curious though.
> 
> The API request itself must be `Future`. Does it therefore matter that inner methods are `FutureOr`?

Both [`convertRequest`](https://github.com/lejard-h/chopper/blob/v4.0.6/chopper/lib/src/interceptor.dart#L78) and [`converResponse`](https://github.com/lejard-h/chopper/blob/v4.0.6/chopper/lib/src/interceptor.dart#L88) of [`Converter`](https://github.com/lejard-h/chopper/blob/v4.0.6/chopper/lib/src/interceptor.dart#L75) are `FutureOr`.

On top of that one can override a `FutureOr` to be sync, I.e. `FutureOr<Response>` can be overridden just as `Response` (what it currently is in the stable release).

techouse:
I had to pin `analyzer` to `>=4.1.0 <4.3.0` because [v4.3.0 deprecates `Element.enclosingElement` and advises to use `Element.enclosingElement2`](https://pub.dev/packages/analyzer/changelog#430)  which is not present in `v4.2.0` and below.

JEuler:
Hm, I'm returning to this thing. As it is breaking we will need to double think it. Personally, I think we can go with your both PR's. 
@stewemetal What do you think? :)

JEuler (approved):

JEuler:
So, I will merge both of your PRs and create a new major update of the Chopper. Thank you for the great work. I think I will do this tomorrow!

PR #349, opened 2022-07-23
Migrate from pedantic to lints ^2.0.0 with lints/recommended.yaml

techouse:
[Dart suggests migrating from package:pedantic](https://github.com/dart-lang/lints#migrating-from-packagepedantic) so I've taken the liberty to re-format the packages and examples.

In the process, I've switched from `pedantic` to `lints: ^2.0.0`, with the recommended lints enables, which [require Dart SDK 2.17.0](https://pub.dev/packages/lints/changelog#200), thus I have bumped the minimum SDK version to `2.17.0`.

If you believe this SDK version is too high I can downgrade to `lints: ^1.0.0` and revert to the previous minimum SDK of `2.12.0`.

As this is more or less a cosmetic PR further reviews and/or scrutiny will be necessary.


techouse:
I have tested the changes in a Flutter app and so far so good, but more tests are needed to make 100% everything works.

JEuler:
Tests are green. And usage is okay. But I'm still thinking that 2.17 maybe too high for this lib... But on the other hand it is better thing to use 2.17 dart. :)

techouse:
@JEuler Agreed. A bump to Dart 2.17 is significant and should require a major version change of this lib.

JEuler (approved):

JEuler:
@techouse Whoops, could you please resolve the conflicts?

techouse:
> @techouse Whoops, could you please resolve the conflicts?

Done 😇

PR #361, opened 2022-09-18
Provide an example using an Isolate Worker Pool with Squadron

techouse:
This is a follow-up on https://github.com/lejard-h/chopper/issues/158#issuecomment-1245991530

Since wiring all this up isn't really that straightforward, I decided to write an example using:

- [json_serializable](https://github.com/google/json_serializable.dart)
- [squadron](https://github.com/d-markey/squadron)
- [squadron_builder](https://github.com/d-markey/squadron_builder)

Essentially all one needs to do is:

1. install the dependencies: `dart pub add squadron squadron_builder`
2. make a service like `example/lib/json_decode_service.dart`
3. run code generation: `dart run build_runner build`
4. wire up `example/lib/json_decode_service.dart` with a custom `JsonConverter`
5. configure and start a `WorkerPool`
6. run HTTP requests like a boss using Isolate Workers 😎

### DIFF
--- a/example/bin/main_json_serializable_squadron_worker_pool.dart
+++ b/example/bin/main_json_serializable_squadron_worker_pool.dart
@@ -1,0 +1,164 @@
+/// This example uses
+/// - https://github.com/google/json_serializable.dart
+/// - https://github.com/d-markey/squadron
+/// - https://github.com/d-markey/squadron_builder
+
+import 'dart:async' show FutureOr;
+import 'dart:convert' show jsonDecode;
+
+import 'package:chopper/chopper.dart';
+import 'package:chopper_example/json_decode_service.dart';
+import 'package:chopper_example/json_serializable.dart';
+import 'package:http/testing.dart';
+import 'package:squadron/squadron.dart';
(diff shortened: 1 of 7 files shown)


JEuler:
Maybe we can add the info somewhere in our documentation? :) 
https://github.com/lejard-h/chopper/blob/develop/faq.md - like here?

techouse:
Aha. Didn't realise that section existed. 😅 Will rather update the markdowns than adding these files.

JEuler:
It is cool to add an example anyway, but we need to point to it in Readme. :)

techouse:
Ok. Will add the info into the readme.

JEuler:
Thank you very much!

techouse:
@JEuler done.

JEuler:
This is so much cool, wow! 🐶 Thank you so much!

JEuler (approved):

techouse:
@JEuler I just checked and the Github pages dox didn't get updated with this content. What needs to be done in order to publish it?

Issue #301, opened 2021-09-28
QueryMap buildUrl 


vanyastaff:
When use QueryMap to get url like example/?filter[age][to]=23&filter[age][from]=14 but I get example/?filter.age.to=23&filter.age.from=14 . :( It make some problem when use some official API. How to solve this problem?

I find code 
if (prefix != null) {
  name = '$prefix.$name';
}

I would more like when use '$prefix[$name]';

techouse:
I'm working on adding a feature that will support that.

techouse:
Fixed in v5.0.1. Closing this issue.

PR #364, opened 2022-10-05
mapToQuery changes

techouse:
This PR will require a bit of discussion:

- I changed how empty strings and `null` are encoded into the query string. 

  I've never seen query parameters taken out if their values were empty or `null` if they were provided to the method encoding the map and I think it's up to the user to clean them up explicitly as most APIs simply ignore them. 
  For comparison, Python's [`urllib.parse.urlencode`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode) employs the same logic.

  ```dart
  // before
  print(mapToQuery({'foo': ''})); // prints ''
  // after
  print(mapToQuery({'foo': ''})); // prints 'foo='
  
  // before
  print(mapToQuery({'foo': null})); // prints ''
  // after
  print(mapToQuery({'foo': null})); // prints 'foo='
  ```

- I changed how empty strings and `null` inside `List`s are encoded into query strings.

  Since these values here are packed into a `List` I'm not sure why they should be encoded, so I removed them (essentially the reverse of the bullet point above which targets standalone values). A query like this is more or less used for filtering by multiple IDs or equivalent where an empty param might actually be detrimental.

  ```dart
  // before
  print(mapToQuery('foo': ['', 'baz', 'etc'])); // prints 'foo=&foo=baz&foo=etc'
  // after
  print(mapToQuery('foo': ['', 'baz', 'etc'])); // prints 'foo=baz&foo=etc'
  
  // before
  print(mapToQuery('foo': [null, 'baz', 'etc'])); // prints 'foo=&foo=baz&foo=etc'
  // after
  print(mapToQuery('foo': [null, 'baz', 'etc'])); // prints 'foo=baz&foo=etc'
  ```

- I made `_normalizeValue` null-safe. Now it will encode `null` to `''` rather than explode violently 😅

- I added the option to use square brackets `[ ]` as an encoded query list and map separator.

  Currently, they are hardcoded to dots `.`, and it's already been expressed that more flexibility is needed https://github.com/lejard-h/chopper/issues/301.
  So I've added an optional `Method` annotation `bool useBrackets`. 

  ```dart
  // using this input in both instances
  final Map<String, dynamic> input = {
     'bar': 'baz',
     'zap': 'abc',
     'etc': {
       'abc': 'def',
       'ghi': 'jkl',
       'mno': {
         'opq': 'rst',
         'uvw': 'xyz',
         'list': ['a', 123, false],
       },
     },
   };

  // the default way using dots as query map separators
  @Get(path: '/map_query_param')
  Future<Response<String>> getUsingMapQueryParam(
    @Query('value') Map<String, dynamic> value,
  );

  // the above produces a URL like this (the line breaks are there just for readability)
  //
  // /test/map_query_param
  // ?value.bar=baz
  // &value.zap=abc
  // &value.etc.abc=def
  // &value.etc.ghi=jkl
  // &value.etc.mno.opq=rst
  // &value.etc.mno.uvw=xyz
  // &value.etc.mno.list=a
  // &value.etc.mno.list=123
  // &value.etc.mno.list=false
  
  // specifying queryMapSeparator: 'brackets'
  @Get(path: '/map_query_param_with_brackets', useBrackets: true)
  Future<Response<String>> getUsingMapQueryParamWithBrackets(
    @Query('value') Map<String, dynamic> value,
  );

  // this produces a URL like this (the line breaks are there just for readability)
  //
  // /test/map_query_param_with_brackets
  // ?value%5Bbar%5D=baz
  // &value%5Bzap%5D=abc
  // &value%5Betc%5D%5Babc%5D=def
  // &value%5Betc%5D%5Bghi%5D=jkl
  // &value%5Betc%5D%5Bmno%5D%5Bopq%5D=rst
  // &value%5Betc%5D%5Bmno%5D%5Buvw%5D=xyz
  // &value%5Betc%5D%5Bmno%5D%5Blist%5D%5B%5D=a
  // &value%5Betc%5D%5Bmno%5D%5Blist%5D%5B%5D=123
  // &value%5Betc%5D%5Bmno%5D%5Blist%5D%5B%5D=false
  //
  // URI decoded this reads as:
  //
  // /test/map_query_param_with_brackets
  // ?value[bar]=baz
  // &value[zap]=abc
  // &value[etc][abc]=def
  // &value[etc][ghi]=jkl
  // &value[etc][mno][opq]=rst
  // &value[etc][mno][uvw]=xyz
  // &value[etc][mno][list][]=a
  // &value[etc][mno][list][]=123
  // &value[etc][mno][list][]=false 
  ```

And of course, I've added a tonne of tests ✅

Check [chopper/tests/utils_test.dart](https://github.com/lejard-h/chopper/pull/364/files#diff-d30cbfe5c31a5a47fc12b9feb9a642bc2d5698d0a2b4da4704e6dfb1c5ab474c) for more examples


JEuler:
Oh my god, this is really cool! Thank you so much, I agree that we should not ignore the null params silently, and should just pass it as it is. 

I saw your request about "hey man release to master" - let's do this after that PR merge. Do you have any plans to add more code here? :)

JEuler (approved):

techouse:
> Do you have any plans to add more code here? :)

Nope :)

techouse:
Done. I added the option to also use brackets in arrays, e.g

`hxxp://path/to/script?foo[]=123&foo[]=456&foo[]=789`

JEuler (approved):

JEuler:
I will prepare the PR to master soon, maybe today, maybe tomorrow morning. Thank you very much for contributing! Need to add you to the author list!

PR #372, opened 2022-10-13
Exclude null query vars by default and add new @Method annotation includeNullQueryVars

### LINKED ISSUE #371: Query with null value included
In the last release 5.0.1 in query, the value is null - URL request was configurated with an empty value.

`/categories?offset=&limit=`

In the previous release, 5.0.0, these queries were not included.

`/categories`

Example:

```dart
@Get(path: 'categories')
  Future<Response<DtoPagination<DtoCategory>>> categories(
    @Header('Request-Service-Location') String service, {
    @Query('offset') int? offset,
    @Query('limit') int? limit,
  });
```

techouse:
Fix #371 

Because some backends do not ignore `null` query parameters I have re-introduced the behaviour of excluding `null` query vars by default. This includes nested `null` query parameters inside maps.

This was the default behaviour prior to `v5.0.1` and the merge of https://github.com/lejard-h/chopper/pull/364.

Additionally, I have added a `@Method` annotation `bool includeNullQueryVars` which, if set to `true`, includes `null` query parameters when encoding them into the URL query. 

- Example with `includeNullQueryVars: false` (default; same behaviour as before `v5.0.1`)

  ```dart
  @Get(path: '/get_data')
  Future<Response<String>> getData({
    @Query('foo') String? foo,
    @Query('bar') String? bar,
    @Query('baz') String? baz,
  });
  
  final response = await service.getData(
    foo: 'foo_val',
    baz: 'baz_val',
  );
  ```
  
  Produces `/test/get_data&foo=foo_val&baz=baz_val`

- Example with `includeNullQueryVars: true` (the current default in `v5.0.1`)

  ```dart
  @Get(
    path: '/get_data', 
    includeNullQueryVars: true,
  )
  Future<Response<String>> getData({
    @Query('foo') String? foo,
    @Query('bar') String? bar,
    @Query('baz') String? baz,
  });
  
  final response = await service.getData(
    foo: 'foo_val',
    baz: 'baz_val',
  );
  ```
  
  Produces `/test/get_data&foo=foo_val&bar=&baz=baz_val`

**NOTE**: Empty `String`s are **always** included. If you do not want them to be included make them nullable, i.e. `String?`

```dart
@Get(path: '/get_data')
Future<Response<String>> getData({
  @Query('foo') String foo = '',
  @Query('bar') String? bar,
  @Query('baz') String? baz,
});

final response = await service.getData(
  bar: null,
  baz: 'baz_val',
);
```

Produces `/test/get_data&foo=&baz=baz_val`

**NOTE 2**: `null` values and empty `String`s inside query `List`s are **always** excluded.

```dart
@Get(path: '/get_data')
Future<Response<String>> getData({
  @Query('foo') List<String?> foo,
});

final response = await service.getData(
  foo: <String?>[
    'value1', 
    '', 
    null, 
    'value4',
  ],
);
```

Produces `/test/get_data&foo=value1&foo=value4`

### DIFF
--- a/chopper_generator/analysis_options.yaml
+++ b/chopper_generator/analysis_options.yaml
@@ -14,7 +14,7 @@ dart_code_metrics:
     cyclomatic-complexity: 20
     number-of-arguments: 4
     maximum-nesting-level: 5
-    number-of-parameters: 6
+    number-of-parameters: 10
     source-lines-of-code: 250
   metrics-exclude:
     - test/**
--- a/chopper/lib/src/annotations.dart
+++ b/chopper/lib/src/annotations.dart
@@ -169,12 +169,16 @@ class Method {
   /// hxxp://path/to/script?user[name]=john&user[surname]=doe&user[age]=21
   final bool useBrackets;
 
+  /// Set to true to truncate all null query variables
+  final bool ignoreNullQueryVars;
+
   const Method(
     this.method, {
     this.optionalBody = false,
     this.path = '',
     this.headers = const {},
     this.useBrackets = false,
+    this.ignoreNullQueryVars = false,
   });
 }
 
@@ -186,6 +190,7 @@ class Get extends Method {
     super.path,
     super.headers,
     super.useBrackets,
+    super.ignoreNullQueryVars,
   }) : super(HttpMethod.Get);
 }
 
@@ -199,6 +204,7 @@ class Post extends Method {
     super.path,
     super.headers,
     super.useBrackets,
+    super.ignoreNullQueryVars,
   }) : super(HttpMethod.Post);
 }
 
@@ -210,6 +216,7 @@ class Delete extends Method {
     super.path,
     super.headers,
     super.useBrackets,
+    super.ignoreNullQueryVars,
   }) : super(HttpMethod.Delete);
 }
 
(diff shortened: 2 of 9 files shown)


luis901101:
Hi and thanks for this fix, I also think that default should be `ignoreNullQueryVars: true` and not otherwise, because as you commented this was the default behavior prior to v5.0.1, so I think it is better to keep the same default to maintain backward compatibility otherwise this will break some existing projects requests, like it's currently happening to me using Java Jooby as Backend.
Not ignoring null query vars as default could be for a new breaking change version like v6.0.0, but I think for v5.x.x the default behavior should remain the same.

techouse:
> Not ignoring null query vars as default could be for a new breaking change version like v6.0.0, but I think for v5.x.x the default behavior should remain the same.

Agreed. I'll amend it.

techouse:
I have renamed `ignoreNullQueryVars` to `includeNullQueryVars`. The default for this setting is `false`, meaning that the behaviour will be reverted to the one prior to `v5.0.1` by default.

**NOTE**: Empty strings are **always** included.

I'll now amend the PR description.

chopper_generator/analysis_options.yaml
```diff
@@ -14,7 +14,7 @@ dart_code_metrics:
     cyclomatic-complexity: 20
     number-of-arguments: 4
     maximum-nesting-level: 5
-    number-of-parameters: 6
+    number-of-parameters: 10
```
JEuler, line 17: Hehe)

JEuler (approved):

JEuler:
LGTM, great work as always, thank you so much! ❤️

PR #375, opened 2022-10-15
[workflow] Upgrade mono_repo to v6.4.1

techouse:
This will resolve the Node 12 deprecation warning

> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/cache, actions/checkout, actions/checkout, actions/cache

Read more [here](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -1,4 +1,4 @@
-# Created with package:mono_repo v6.0.0
+# Created with package:mono_repo v6.4.1
 name: Dart CI
 on:
   push:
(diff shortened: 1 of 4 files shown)


techouse:
Looks like Githiub still complains about `save-state`, however, this is the latest [mono_repo](https://pub.dev/packages/mono_repo) at the moment

> The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

JEuler (approved):

PR #381, opened 2022-10-18
MonoRepo update

JEuler:


JEuler:
I don't know if it is needed here but let it be I think. :)

techouse (approved):

techouse:
It is, because otherwise the Github actions will start failing in the future.

Issue #304, opened 2021-10-12
[Feature]: Can we replace the String based path with Uri() instead?


masseelch:
Currently: 
```dart
 final r = await client.post(
          '/auth/token?cred=something%20with%20spaces',
          body: {
            'email': _email,
            'password': _password,
          },
        );
```
I propose:
```dart
 final r = await client.post(
          Uri(path: '/auth/token', queryParameters: {'cred': 'something with spaces'}),,
          body: {
            'email': _email,
            'password': _password,
          },
        );
```

Underlying http client does use Uri too. I can provide a PR if you agree with this proposal.

JEuler:
Seems nice, could you please provide the PR?

JEuler:
Cool, thank you very much!

JEuler:
@masseelch Any luck with that? :) Hi!

masseelch:
Ah, ooopsi. I completly forgot. Was on vacation and havent touched the project since then. I will try my luck tomorrow :-)

JEuler:
Thank you very much! 🍺

masseelch:
Buried deep in work. But I got some spare time tomorow :-)

JEuler:
All is okay, I'm also totally in work. :)

masseelch:
Had a look yesterday and the changes are quite huge. I won't find time to do the changes in the foreseeable future. And this is not a critical bug, just a nice to have. 

I'll leave this open, i might find time someday. Or somebody else is picking this up.

PR #383, opened 2022-10-21
[Feature] Replace the String based path with Uri

Guldem:
I took the liberty of replacing the `String` based url path in the client calls with an `Uri` as request in the #304 feature request.

This is now more in line with the underlying `Http.Client`. 

Any feedback is welcome :)

### DIFF
--- a/chopper/lib/src/base.dart
+++ b/chopper/lib/src/base.dart
@@ -341,35 +341,32 @@ class ChopperClient {
 
   /// Makes a HTTP GET request using the [send] function.
   Future<Response<BodyType>> get<BodyType, InnerType>(
-    String url, {
+    Uri url, {
     Map<String, String> headers = const {},
-    Map<String, dynamic> parameters = const {},
     String? baseUrl,
     dynamic body,
   }) =>
       send<BodyType, InnerType>(
-        Request(
+        Request.uri(
           HttpMethod.Get,
           url,
           baseUrl ?? this.baseUrl,
           body: body,
           headers: headers,
-          parameters: parameters,
         ),
       );
 
   /// Makes a HTTP POST request using the [send] function
   Future<Response<BodyType>> post<BodyType, InnerType>(
-    String url, {
+    Uri url, {
     dynamic body,
     List<PartValue> parts = const [],
     Map<String, String> headers = const {},
-    Map<String, dynamic> parameters = const {},
     bool multipart = false,
     String? baseUrl,
   }) =>
       send<BodyType, InnerType>(
-        Request(
+        Request.uri(
           HttpMethod.Post,
           url,
           baseUrl ?? this.baseUrl,
@@ -377,22 +374,20 @@ class ChopperClient {
           parts: parts,
           headers: headers,
           multipart: multipart,
-          parameters: parameters,
         ),
       );
 
   /// Makes a HTTP PUT request using the [send] function.
   Future<Response<BodyType>> put<BodyType, InnerType>(
-    String url, {
+    Uri url, {
     dynamic body,
     List<PartValue> parts = const [],
     Map<String, String> headers = const {},
-    Map<String, dynamic> parameters = const {},
     bool multipart = false,
     String? baseUrl,
   }) =>
       send<BodyType, InnerType>(
-        Request(
+        Request.uri(
           HttpMethod.Put,
           url,
           baseUrl ?? this.baseUrl,
@@ -400,22 +395,20 @@ class ChopperClient {
           parts: parts,
           headers: headers,
           multipart: multipart,
-          parameters: parameters,
         ),
       );
 
   /// Makes a HTTP PATCH request using the [send] function.
   Future<Response<BodyType>> patch<BodyType, InnerType>(
-    String url, {
+    Uri url, {
     dynamic body,
     List<PartValue> parts = const [],
     Map<String, String> headers = const {},
-    Map<String, dynamic> parameters = const {},
     bool multipart = false,
     String? baseUrl,
   }) =>
       send<BodyType, InnerType>(
-        Request(
+        Request.uri(
           HttpMethod.Patch,
           url,
           baseUrl ?? this.baseUrl,
@@ -423,58 +416,51 @@ class ChopperClient {
           parts: parts,
           headers: headers,
           multipart: multipart,
-          parameters: parameters,
         ),
       );
 
   /// Makes a HTTP DELETE request using the [send] function.
   Future<Response<BodyType>> delete<BodyType, InnerType>(
-    String url, {
+    Uri url, {
     Map<String, String> headers = const {},
-    Map<String, dynamic> parameters = const {},
     String? baseUrl,
   }) =>
       send<BodyType, InnerType>(
-        Request(
+        Request.uri(
           HttpMethod.Delete,
           url,
           baseUrl ?? this.baseUrl,
           headers: headers,
-          parameters: parameters,
         ),
       );
 
   /// Makes a HTTP HEAD request using the [send] function.
   Future<Response<BodyType>> head<BodyType, InnerType>(
-    String url, {
+    Uri url, {
     Map<String, String> headers = const {},
-    Map<String, dynamic> parameters = const {},
     String? baseUrl,
   }) =>
       send<BodyType, InnerType>(
-        Request(
+        Request.uri(
           HttpMethod.Head,
           url,
           baseUrl ?? this.baseUrl,
           headers: headers,
-          parameters: parameters,
         ),
       );
 
   /// Makes a HTTP OPTIONS request using the [send] function.
   Future<Response<BodyType>> options<BodyType, InnerType>(
-    String url, {
+    Uri url, {
     Map<String, String> headers = const {},
-    Map<String, dynamic> parameters = const {},
     String? baseUrl,
   }) =>
       send<BodyType, InnerType>(
-        Request(
+        Request.uri(
           HttpMethod.Options,
           url,
           baseUrl ?? this.baseUrl,
           headers: headers,
-          parameters: parameters,
         ),
       );
 
--- a/chopper/lib/src/base.dart
+++ b/chopper/lib/src/base.dart
@@ -123,7 +123,7 @@ class ChopperClient {
             baseUrl == null || !baseUrl.hasQuery,
             'baseUrl should not contain query parameters.'
             'Use a request interceptor to add default query parameters'),
-        baseUrl = baseUrl ?? Uri.parse(''),
+        baseUrl = baseUrl ?? Uri(),
         httpClient = client ?? http.Client(),
         _clientIsInternal = client == null {
     if (!interceptors.every(_isAnInterceptor)) {
--- a/chopper/lib/src/base.dart
+++ b/chopper/lib/src/base.dart
@@ -29,7 +29,7 @@ final List<Type> allowedInterceptorsType = [
 class ChopperClient {
   /// Base URL of each request of the registered services.
   /// E.g., the hostname of your service.
-  final String baseUrl;
+  final Uri baseUrl;
 
   /// The [http.Client] used to make network calls.
   final http.Client httpClient;
@@ -60,6 +60,7 @@ class ChopperClient {
   /// The base URL of each request of the registered services can be defined
   /// with the [baseUrl] parameter.
   ///  E.g., the hostname of your service.
+  ///  If not provided, a empty default [Uri] will be used.
   ///
   /// A custom HTTP client can be passed as the [client] parameter to be used
   /// with the created [ChopperClient].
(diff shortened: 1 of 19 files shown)


JEuler:
Hi! Cool, thx! Please fix the formatter issues! :)

Guldem:
Woeps other default line length configured :) But reformatted the files now

chopper/lib/src/request.dart
```diff
@@ -45,23 +45,27 @@ class Request extends http.BaseRequest {
...
     this.multipart = false,
     this.parts = const [],
     this.useBrackets = false,
     this.includeNullQueryVars = false,
-  })  : origin = url.origin,
+  })  : origin = url.isScheme('HTTP') || url.isScheme('HTTPS')
+            ? url.origin
+            : baseUrl,
         path = url.path,
-        parameters = {...url.queryParametersAll, ...?parameters},
```
techouse, line 58: Hmm, I kept this here specifically for edge cases when one wants to add additional parameters to the Uri without messing with the Uri itself.

chopper/test/request_test.dart
```diff
@@ -92,53 +92,55 @@ void main() {
...
+          '',
         ).url,
         equals(Uri.parse('https://foo/bar?lorem=ipsum&dolor=123')),
       );
 
       expect(
         Request.uri(
           'GET',
-          Uri.parse(
-            'https://foo/bar?first=sit&second=amet&first_list=a&first_list=b',
```
techouse, line 135: This test here for example merges a Uri with existing parameters with an additional Map from the `Request.uri` constructor's `parameters` parameter.

techouse:
Nice one 😎 

Was planning to do this eventually after I did https://github.com/lejard-h/chopper/pull/370 but you beat me to it 😂

techouse:
Also, since you're already doing this it might make sense to also update the generator to use the `Request.uri` constructor instead of the default one that still uses `String`s.

chopper/lib/src/request.dart
Guldem: I think it could be confusing having multiple ways of supplying parameters. Of course it wouldn't hurt to keep both. But even if someone doesn't want to mess with the orginal Uri,  one could use `Uri.replace` which creates a new instance and leaving the old intact.  An egde case like that probably isn't hard to work around.

But maybe I don't really see the edge case right.

Guldem:
> Also, since you're already doing this it might make sense to also update the generator to use the `Request.uri` constructor instead of the default one that still uses `String`s.

First I started to do that but wasn't sure if that is preferable. Also if the generator uses `Uri` I think the `Request` could be fully `Uri` based and it would be able to remove the `Request.uri` constructor.

techouse:
> First I started to do that but wasn't sure if that is preferable. Also if the generator uses `Uri` I think the `Request` could be fully `Uri` based and it would be able to remove the `Request.uri` constructor.

Yep, I wrote the `Request.uri` constructor specifically for the purpose of transitioning the package to a fully `Uri` based `Request`. Once that's done it can be safely removed and a new major version released.

chopper/lib/src/request.dart
techouse: The [Uri.queryParameters](https://github.com/dart-lang/sdk/blob/stable/sdk/lib/core/uri.dart#L212) only accepts a `Map` containing `String` or `Iterable<String>`. 

```dart
// this compiles fine
final Uri url = Uri.https(
  'chopper.dev',
  '/test',
  {
    'foo': 'bar',
    'foo_list': [
      'one',
      'two',
      'three',
    ],
  },
);
```

If you try to use a `Map<String, String>` it will throw an exception

`type '_InternalLinkedHashMap<String, String>' is not a subtype of type 'Iterable<dynamic>'`

```dart
// this does not compile
final Uri url= Uri.https(
  'chopper.dev',
  '/test2',
  {
    'foo': 'bar',
    'foo_list': [
      'one',
      'two',
      'three',
    ],
    'user': {
      'name': 'john',
      'surname': 'doe',
    },
  },
);
```

Chopper's `Request.parameters`, however, will accept a `Map<String, dynamic>`.

```dart
test('works', () {
  expect(
    Request.uri(
      'GET',
      Uri.parse(
        'https://chopper.dev/test3',
      ),
      parameters: {
        'foo': 'bar',
        'foo_list': [
          'one',
          'two',
          'three',
        ],
        'user': {
          'name': 'john',
          'surname': 'doe',
        },
      },
    ).url.toString(),
    equals(
      'https://chopper.dev/test3?'
      'foo=bar'
      '&foo_list=one'
      '&foo_list=two'
      '&foo_list=three'
      '&user.name=john'
      '&user.surname=doe',
    ),
  );
});
```

Chopper has had this [feature](https://github.com/lejard-h/chopper/blob/develop/chopper/lib/src/utils.dart#L90) for a while and people might rely on it.

On top of that, the getters [Uri.queryParameters](https://github.com/dart-lang/sdk/blob/stable/sdk/lib/core/uri.dart#L4521) and [Uri.queryParametersAll](https://github.com/dart-lang/sdk/blob/stable/sdk/lib/core/uri.dart#L4526) both parse the already encoded query string to get the parameters back into a `Map<String, dynamic>` which means that the original `Map<String, dynamic>` that it passed to the `Uri` constructor is not preserved.

That's why [I merged both](https://github.com/lejard-h/chopper/blob/develop/chopper/lib/src/request.dart#L58) before sending the final `parameters` to [buildUri](https://github.com/lejard-h/chopper/blob/develop/chopper/lib/src/request.dart#L102).

JEuler:
Yeah, totally agree, that a fully URI approach is the preferred option. So, probably this work can be done in this PR. 👍

chopper/lib/src/request.dart
Guldem: Aah makes more sense now. I will revert the removal of the parameters for now.

techouse: Cool. 😎 While you're at it, please dump that test case above into [chopper/test/request_test.dart](https://github.com/lejard-h/chopper/blob/develop/chopper/test/request_test.dart). 😇

Guldem:
Did a bunch of refactors to use `Uri` throughout the code base. But it's mostly replacing the `String`s with `Uri` at the moment. Probably I have to rethink some stuff also haven't added any specific tests to make sure everything works. Although the current available tests succeed 😄

Any feedback is welcome ofc 😄

techouse:
> Any feedback is welcome ofc 😄

Take your time dude. This is a big job 🤓

P.S. You might wanna convert it to a Draft PR until you think it's ready. I'll take a look again tomorrow after work.

chopper/lib/src/request.dart
```diff
@@ -100,17 +71,17 @@ class Request extends http.BaseRequest {
...
     bool useBrackets = false,
     bool includeNullQueryVars = false,
   }) {
     // If the request's url is already a fully qualified URL, we can use it
     // as-is and ignore the baseUrl.
-    final Uri uri = url.startsWith('http://') || url.startsWith('https://')
-        ? Uri.parse(url)
-        : Uri.parse('${baseUrl.strip('/')}/${url.leftStrip('/')}');
+    final Uri uri = url.isScheme('HTTP') || url.isScheme('HTTPS')
+        ? _mergeUri(url, baseUrl, mergePaths: false)
```
Guldem, line 83: I added this with the idea of making it possible to preserve query parameters of the `baseUrl` (in the `Uri`). But I'm wondering if that is ideal. 

This makes it impossible the remove or override the query parameters provided from `baseUrl`. Only in the client calls (`get`, `post` etc) it's possible to override the `baseUrl` and also override any query parameter. But the `baseUrl`.

I kinda liked the idea of preserving the information but I can also imaging a lot of annoyances when its not preferred. What do you think @techouse?

techouse: Hmm, to me putting query parameters into the `baseUrl` is just a crazy idea and more or less a hack.

Maybe you should add a validator that validates the `baseUrl` and throws an `Exception` when it detects query parameters in it. Even a simple `String.contains('?')` might do the trick.

Guldem: I guess its also kinda hacky besides its cleaner to provide a interceptor to add default query parameter if that would be needed.

techouse:
I'll review this tomorrow. 🤓

chopper/lib/src/base.dart
```diff
@@ -111,14 +112,19 @@ class ChopperClient {
...
     http.Client? client,
     Iterable interceptors = const [],
     this.authenticator,
     this.converter,
     this.errorConverter,
     Iterable<ChopperService> services = const [],
-  })  : httpClient = client ?? http.Client(),
+  })  : assert(
+            baseUrl == null || !baseUrl.hasQuery,
+            'baseUrl should not contain query parameters.'
```
techouse, line 124: Ha! Smart! I love it! 😎

chopper/lib/src/request.dart
```diff
@@ -1,14 +1,14 @@
...
+import 'package:chopper/chopper.dart';
 import 'package:chopper/src/utils.dart';
 import 'package:http/http.dart' as http;
 import 'package:meta/meta.dart';
 
 /// This class represents an HTTP request that can be made with Chopper.
 class Request extends http.BaseRequest {
-  final String path;
-  final String origin;
+  final Uri uri;
```
techouse, line 10: Hmm, why did you create another `Uri` field? Was the parent `BaseRequest.url` not good enough? I don't see it being used anywhere else besides within the class, so I don't think it's needed. Simply use the parent `BaseRequest.url`.

chopper/lib/src/request.dart
```diff
@@ -1,14 +1,14 @@
 import 'dart:async';
 
-import 'package:chopper/src/extensions.dart';
+import 'package:chopper/chopper.dart';
```
techouse, line 3: Why? Only the extension is needed. I don't see a point in importing the whole package for it.

chopper/lib/src/request.dart
```diff
@@ -100,27 +77,44 @@ class Request extends http.BaseRequest {
...
     // as-is and ignore the baseUrl.
-    final Uri uri = url.startsWith('http://') || url.startsWith('https://')
-        ? Uri.parse(url)
-        : Uri.parse('${baseUrl.strip('/')}/${url.leftStrip('/')}');
+    final Uri uri = url.isScheme('HTTP') || url.isScheme('HTTPS')
+        ? url
+        : _mergeUri(baseUrl, url);
+
+    // Check if parameter also has all the queryParameters from the url (not the merged uri)
+    final parametersContainsUriQuery = parameters.keys
```
techouse, line 93: Please add the `bool` type annotation.

chopper/lib/src/request.dart
```diff
@@ -100,27 +77,44 @@ class Request extends http.BaseRequest {
...
+        .every((element) => url.queryParametersAll.keys.contains(element));
+    final allParameters = parametersContainsUriQuery
```
techouse, line 95: Please add the `Map<String, dynamic>` type annotation.

chopper_generator/lib/src/generator.dart
```diff
@@ -475,21 +475,26 @@ class ChopperGenerator extends GeneratorForAnnotation<chopper.ChopperApi> {
...
-        return literal('$baseUrl/$path');
+        return _generateUri('$baseUrl/$path');
       }
 
-      return literal('$baseUrl$path');
+      return _generateUri('$baseUrl$path');
     }
   }
 
+  Expression _generateUri(String url) => refer('Uri').newInstanceNamed(
```
techouse, line 493: Smart love it!

techouse:
Lovely! Some really smart decisions there! 😎

Please check my comments. I'll check back again tomorrow.

chopper/lib/src/request.dart
Guldem: Probably auto import from IDE😓

chopper/lib/src/request.dart
Guldem: Main reason was the `copyWith` of the `Request`. When creating a copy of `Request` the merged `Uri url` would be used for `uri` (or both `baseUri`) if not overridden. Having a merged `uri` with `http/https` would always go in favor of the `baseUri` and making it impossible to override the `baseUri` without also providing a `uri`.

Another solution could be removing `uri` & `baseUri` from `Request` & `copyWith` in favour of `url`. And making a factory constructor with `uri` & `baseUri` which get merged. But might introduces the query parameter copying/syncing a bit more complicated.

techouse: I see. Tricky one, however, I'd actually prefer that `factory` constructor and clean stuff up even more, like you said.

@JEuler your thoughts?

techouse (approved):

techouse:
Great work. Just gotta wait for @JEuler to throw an eye on it :)

chopper/lib/src/request.dart
JEuler: Hm, Let's go with what we have now, it is okay to do a bit of refactoring later. Great work, thank you for contribution @Guldem @techouse

JEuler:
I will merge it, and the "factory refactor" can be made later. Also, let's "accumulate" more changes before the release! 👍 
LGTM!

PR #386, opened 2022-11-20
:memo: Add Authenticator example

techouse:
I've seen at least 3 issues related to setting up the [Authenticator](https://github.com/lejard-h/chopper/blob/develop/chopper/lib/src/authenticator.dart).

Digging out an example from https://github.com/lejard-h/chopper/pull/198#issuecomment-761781430 is just too cumbersome, so I've added an entry into the FAQ.


JEuler (approved):

JEuler:
Good work, Thank you !

PR #399, opened 2023-01-04
Configurable HttpLoggingInterceptor

Guldem:
Updated the `HttpLoggingInterceptor` so users can specify a more fine grained logging level for there needs.

Inspired by the `HttpLoggingInterceptor` configuration of OkHttp.

### DIFF
--- a/chopper/lib/src/http_logging_interceptor.dart
+++ b/chopper/lib/src/http_logging_interceptor.dart
@@ -1,0 +1,171 @@
+import 'dart:async';
+
+import 'package:chopper/src/interceptor.dart';
+import 'package:chopper/src/request.dart';
+import 'package:chopper/src/response.dart';
+import 'package:chopper/src/utils.dart';
+import 'package:http/http.dart' as http;
+import 'package:meta/meta.dart';
+
+enum Level {
+  /// No logs.
+  none,
+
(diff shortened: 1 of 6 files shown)


chopper/lib/src/http_logging_interceptor.dart
```diff
@@ -0,0 +1,172 @@
...
+  final bool _logHeaders;
+
+  @override
+  FutureOr<Request> onRequest(Request request) async {
+    if (level == Level.none) return request;
+    final http.BaseRequest base = await request.toBaseRequest();
+
+    String startRequestMessage = '--> ${base.method} ${base.url.toString()}';
+    String bodyMessage = '';
+    if (base is http.Request) {
```
techouse, line 89: You could join the `if` conditions to something like
```dart
if (base is http.Request) {
  if (base.body.isNotEmpty) {
    bodyMessage = base.body;

    if (!_logHeaders) {
      startRequestMessage += ' (${base.bodyBytes.length}-byte body)';
    }
  }
}
```

techouse (changes requested):

techouse (approved):
LGTM

JEuler (approved):
Cool, many thanks!

PR #400, opened 2023-01-06
Made apply headers field name case insensitive.

### LINKED ISSUE #398: Content-Type for JsonConverter request  is case-sensitive causing applyingHeader to always override
# Problem
When using the default `JsonConverter` for requests that contain a header with `Content-Type`, the value is always overridden with `content-type: application/json`, even though `override: false`. The issue is due to `JsonConverter` always expecting `content-type` to be in lowercase.

# Steps to Reproduce
1. Using the default `JsonConverter`. Make a request with headers containing  e.g `Content-Type: application/vnd.api+json`
2. Inspect or log the final outgoing header, which will instead contain `content-type: application/json`.

# Workaround
Use lowercase `content-type` instead of `Content-Type`

Guldem:
Fixes #398

`applyHeaders` would override headers with different cases. But the header field names should be case-insensitive according to the [RFC 7540](https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2).

This MR makes applying headers case insensitive. Also added bunch of tests for verifying `applyHeaders`.

### DIFF
--- a/chopper/lib/src/utils.dart
+++ b/chopper/lib/src/utils.dart
@@ -1,3 +1,5 @@
+import 'dart:collection';
+
 import 'package:chopper/chopper.dart';
 import 'package:logging/logging.dart';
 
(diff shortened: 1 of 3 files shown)


techouse (approved):
LGTM

techouse:
Thanks for this! 😇

JEuler:
Sorry, moving around the world now guys! LGTM! Thank you so much!

JEuler (approved):

PR #403, opened 2023-01-09
:bulb: Clean up code in comments

techouse:
I noticed that some code inside comment blocks (`///`) was annotated to start with ` ```dart ` but didn't end with ` ``` `.

As a result, the code wasn't properly highlighted on Github, for example [here](https://github.com/lejard-h/chopper/blob/develop/chopper/lib/src/annotations.dart#L330).

**NOTE** this is a cosmetic fix only. No code was changed, only comments.


JEuler (approved):

JEuler:
Hooray!

PR #407, opened 2023-02-06
Upgrade of chopper_built_value and chopper version

JEuler:


JEuler:
Whoops I forget we have more changes / need to update the changelog and version accordingly. Will do it tomorrow.

JEuler:
As I understand, nothing need to be updated in terms of version.

techouse (approved):

PR #430, opened 2023-05-25
Adding follow redirects to toHttpRequest

Joran-Dob:
Currently `followRedirects` are not passed to `http.Request` in this method `toHttpRequest`. This causes setting `followRedirects` to `false` for example in a interceptor to not be passed http client.


techouse (approved):

techouse:
Thanx!

Joran-Dob:
> Thanx!

Thanks for the great package, small question. Any idea when this change will be in a release? 😄

techouse:
> Any idea when this change will be in a release? 😄

Soon, I hope 😅

techouse:
> Any idea when this change will be in a release? 😄

@Joran-Dob the change was released in v6.1.3

Issue #438, opened 2023-05-28
List<int> is not sent as string in multipart request.


Simouche:
I have this multipart request:
```  @Post(path: "/", headers: {"content-type": "multipart/form-data"})
  @multipart
  Future<Response<Homework>> create_(
    @Part("teacher") int teacher,
    @Part("title") String title,
    @Part("description") String description,
    @Part("subject") int subject,
    @Part("classes") List<int> classes,
    @Part("dueDate") String dueDate,
    // @PartFile("homeworkFiles") List<PartValueFile<String>> homeworkFiles,
  );
```

the classes part, is being sent as a string, and causing the backend to respond with http 400 with this error:
```
{"classes":["Incorrect type. Expected pk value, received str."],"homeworkFiles":["This field is required."]}
```
anyway to fix this issue?

Also, how to send multiple files under the same field name? uncommenting the last line makes it fail (without throwing any errors).

techouse:
Hey 👋 

You might want to look at [this method](https://github.com/lejard-h/chopper/blob/develop/chopper/lib/src/request.dart#L161) for clues.

Would be great if you can come up with a [test case here](https://github.com/lejard-h/chopper/blob/develop/chopper/test/multipart_test.dart).

techouse:
@Simouche can you check if this fixes it for you? https://github.com/lejard-h/chopper/pull/439

Simouche:
@techouse how do i test it? i tried referencing it like this:
```
  chopper:
    git:
      url: https://github.com/techouse/chopper.git
      ref: fix/issue-438-multipart-list
```

and i had this error:
```
Could not find a file named "pubspec.yaml" in https://github.com/techouse/chopper.git e70429782820caa346cc299909eec0a5dc1f28fb.
pub get failed
command: "C:\flutter\bin\cache\dart-sdk\bin\dart __deprecated_pub --directory . get --example"
pub env: {
  "FLUTTER_ROOT": "C:\flutter",
  "PUB_ENVIRONMENT": "flutter_cli:get",
  "PUB_CACHE": "C:\Users\pc\AppData\Local\Pub\Cache",
}
exit code: 1

```

techouse:
> chopper:
>     git:
>       url: https://github.com/techouse/chopper.git
>       ref: fix/issue-438-multipart-list

You need to also specify the directory.

Simouche:
yeah right, i did that and i got this now:
```
[...]

Issue #316, opened 2021-12-28
Sending list<int> in a multipart request


alihabbash:
I'm trying to send a list (it's not a file) within a multipart request
The api gives me an error `the days must be an array`
My way of doing it like this:
```
  @Post(path: Urls.REGISTER)
  @Multipart()
  Future<Response> register({
    @Part(ApiConst.PHONE_NUM_FIELD) required String phone,
    @Part(ApiConst.CONFIRM_CODE_FIELD) required String confirmCode,
    @Part(ApiConst.FIRST_NAME_FIELD) required String firstName,
    @Part(ApiConst.LAST_NAME_FIELD) required String lastName,
    @Part(ApiConst.BIRTH_DATE_FIELD) required String birthDate,
    @Part(ApiConst.GENDER_FIELD) required String gender,
    @Part(ApiConst.CITY_ID_FIELD) required int cityId,
    @Part(ApiConst.NATIONAL_ID_FIELD) required String nationalNum,
    @Part(ApiConst.VEHICLE_ID_FIELD) required int vehicleId,
    @Part(ApiConst.VEHICLE_NUMBER_FIELD) required String vehicleNumber,
    @Part(ApiConst.MADE_YEAR_FIELD) required String madeYear,
    @Part(ApiConst.VEHICLE_COLOR_FIELD) required String vehicleColor,
    @Part(ApiConst.VEHICLE_SUB_BRAND_FIELD) required String vehicleSubBrand,
    @Part(ApiConst.VEHICLE_CLASS_ID_FIELD) required int vehicleClassId,
    @Part(ApiConst.SECOND_WORK_CITY_ID_FIELD) int? secondWorkCity,
    @Part(ApiConst.WORK_DAYS_FIELD) List<int>? workDaysIds,
    @PartFile(ApiConst.VEHICLE_LICENSE_FIELD)
        required String vehicleLicenseFilePath,
    @PartFile(ApiConst.DRIVER_LICENSE_FIELD)
        required String driverLicenseFilePath,
    @PartFile(ApiConst.VEHICLE_REAR_FIELD) required String vehicleRearFilePath,
    @PartFile(ApiConst.VEHICLE_FRONT_FIELD)
        required String vehicleFrontFilePath,
    @PartFile(ApiConst.NON_CONVICTION_FIELD)
        required String nonConvictionFilePath,
  });
```

Simouche:
how did you fix this?

Issue #77, opened 2019-08-19
Multiple multipartFiles with same field name


josephchenghmlet:
Is there an alternative to pass multiple multipartFiles with same field name?
This issue was issued from [http #277](https://github.com/dart-lang/http/issues/277).

humazed:
In chopper, you can send multiple files with the same name,
example:

```dart
  @Post(path: "store/orders")
  @multipart
  Future<Response<Order>> createTripOrder(
    @Part("trip") int trip,
    @PartFile("images[]") List<MultipartFile> images,
  );

```

josephchenghmlet:
Hi @humazed, thanks! I'll try your answer. However, do you have suggestion for #75?

Purvik:
@humazed  If you can answer this one I am getting trouble with single file upload
https://github.com/lejard-h/chopper/issues/129

PR #439, opened 2023-05-29
:bug: fix Multipart for List<int> and List<String>

techouse:
Adds the ability to do something like

```dart
@Post(path: 'multipart_list')
@multipart
Future<Response> postMultipartList({
  @Part('ints') required List<int> ints,
  @Part('doubles') required List<double> doubles,
  @Part('nums') required List<num> nums,
  @Part('strings') required List<String> strings,
});

final Response res = await postMultipartList(
  ints: <int>[1, 2, 3],
  doubles: <double>[1.23, -1.23, 0.0, 0.12324, 3 / 4],
  nums: <num>[1.23443534678, 0.00000000001, -34251, 0.0, 3 / 4],
  strings: <String>['lorem', 'ipsum', 'dolor', '''r237tw78re ei[04o2 ]de[qwlr;,mgrrt9ie0owp[ld;s,a.vfe[plre'q/sd;poeßšđčćž'''],
);
```

which will produce this payload (cURL form just for brevity)

```bash
curl -v -X POST \
	-f 'ints[0]: 1' \
	-f 'ints[1]: 2' \
	-f 'ints[2]: 3' \
	-f 'doubles[0]: 1.23' \
	-f 'doubles[1]: -1.23' \
	-f 'doubles[2]: 0.0' \
	-f 'doubles[3]: 0.12324' \
	-f 'doubles[4]: 0.75' \
	-f 'nums[0]: 1.23443534678' \
	-f 'nums[1]: 0.00000000001' \
	-f 'nums[2]: -34251' \
	-f 'nums[3]: 0.0' \
	-f 'nums[4]: 0.75' \
	-f 'strings[0]: lorem' \
	-f 'strings[1]: ipsum' \
	-f 'strings[2]: dolor' \
	-f "strings[3]: r237tw78re ei[04o2 ]de[qwlr;,mgrrt9ie0owp[ld;s,a.vfe[plre'q/sd;poeßšđčćž" \
	https://foo.dev/multipart_list"
```

The gist of this is to use bracket notation (`field[0]`, `field[1]`, `field[2]`, ...) to overcome the limitation of [`MultipartRequest.fields`](https://pub.dev/documentation/http/latest/http/MultipartRequest/fields.html) which is a `Map<String, String>` and as such can not have repeated keys.

This limitation has been discussed before in https://github.com/dart-lang/http/issues/277 and https://github.com/dart-lang/http/issues/24.

[The solution I've implemented](https://github.com/lejard-h/chopper/pull/439/files#diff-2dc27a5036ef0c21dd7610b955335660b980abb03efbe31bd900159aa2fb4f6aR190) is exactly the same as suggested here https://github.com/dart-lang/http/issues/277#issuecomment-496694756

---

Addresses #77, #316 and #438


JEuler:
I've read your discussion on the issue and looked at the PR. As I understand, nothing can be done better, because Chopper utilizes the http package. 

So, we can, of course, provide something like an option to offer the capability to join list of ints to one array... but I don't feel like we should because it is easy to do by using "stringification". So, LGTM.

JEuler (approved):

techouse:
> So, we can, of course, provide something like an option to offer the capability to join list of ints to one array... but I don't feel like we should because it is easy to do by using "stringification".

Well yes, in the case of ints it's easy, but as soon as you have strings or something more complex it explodes. 🙈

Issue #443, opened 2023-06-18
It's not possible to put generated files to a separate directory


victor-semenovich-dev:
The generated files in my project should be placed to the separate directory "/lib/generated". It works for the source_gen package, but not for the chopper_generator. Here is my build.yaml file:

```
targets:
  $default:
    builders:
      source_gen|combining_builder:
        options:
          build_extensions:
            '^lib/{{}}.dart': 'lib/generated/{{}}.g.dart'
      chopper_generator:
        generate_for:
          - "lib/data/remote/service/**.dart"
        options:
          build_extensions:
            '^lib/{{}}.dart': 'lib/generated/{{}}.chopper.dart'
```

Is it possible to configure the build.yaml to put the generated files to the separate directory?

techouse:
If my memory serves me right this isn't yet supported.

techouse:
@victor-semenovich-dev I have made a PR here #444 

Would you mind checking out the repo and testing against your config?

Add this to your `pubspec.yml` to use the PR directly in your project

```yaml
dependency_overrides:
  chopper_generator:
    git:
      url: https://github.com/techouse/chopper.git
      ref: feat/issue-443-build-extensions
      path: chopper_generator
```

victor-semenovich-dev:
@techouse Works like a charm, thank you!

techouse:
Once #444 hets merged I'll go ahead an release a new version, however, it could take up to a few days. In the meantime just use the PR branch.

techouse:
Released in `chopper_generator` **v6.0.2**

PR #444, opened 2023-06-18
:sparkles: Generating files in different directories

techouse:
Give the user the option for generating files in different directories, just as [described here](https://pub.dev/packages/source_gen#generating-files-in-different-directories).

Example of a `build.yml` that will put `*.chopper.dart` files in a custom `lib/generated` directory

```yaml
targets:
  $default:
    builders:
      chopper_generator:
        options:
          build_extensions:
            '^lib/{{}}.dart': 'lib/generated/{{}}.chopper.dart'
```

---

Addresses #443


JEuler (approved):

JEuler:
Very cool!

PR #448, opened 2023-06-19
:recycle: refactor chopper_generator

techouse:
# :recycle: refactor chopper_generator

- [x] simplify library export
- [x] extract `PartBuilder` into its own file
- [x] extract constant variables into an `enum` 
- [x] extract helper methods into a `Utils` class
- [x] use a `const` constructor
- [x] make all  methods `static`
- [x] ensure all immutable variables are `final`
- [x] simplify syntax 
- [x] add API documentation 
- [x] add [build_verify](https://github.com/kevmoo/build_verify) test
- [x] add [build_verify](https://github.com/kevmoo/build_verify) test to chopper as well because it also uses generated files in the test suite
- [x] update README

### DIFF
--- a/chopper/test/ensure_build_test.dart
+++ b/chopper/test/ensure_build_test.dart
@@ -7,7 +7,7 @@ void main() {
   test(
     'ensure_build',
     () => expectBuildClean(
-      packageRelativeDirectory: 'chopper_generator',
+      packageRelativeDirectory: 'chopper',
       gitDiffPathArguments: [
         'test/test_service.chopper.dart',
       ],
--- a/chopper_generator/lib/src/generator.dart
+++ b/chopper_generator/lib/src/generator.dart
@@ -16,6 +16,8 @@ import 'package:logging/logging.dart';
 import 'package:source_gen/source_gen.dart';
 
 class ChopperGenerator extends GeneratorForAnnotation<chopper.ChopperApi> {
+  const ChopperGenerator();
+
   static final Logger _logger = Logger('Chopper Generator');
 
   @override
--- a/chopper_generator/lib/src/generator.dart
+++ b/chopper_generator/lib/src/generator.dart
@@ -25,10 +25,10 @@ class ChopperGenerator extends GeneratorForAnnotation<chopper.ChopperApi> {
     BuildStep buildStep,
   ) {
     if (element is! ClassElement) {
-      final String friendlyName = element.displayName;
       throw InvalidGenerationSourceError(
-        'Generator cannot target `$friendlyName`.',
-        todo: 'Remove the [ChopperApi] annotation from `$friendlyName`.',
+        'Generator cannot target `${element.displayName}`.',
+        todo:
+            'Remove the [ChopperApi] annotation from `${element.displayName}`.',
       );
     }
 
--- a/chopper_generator/lib/src/generator.dart
+++ b/chopper_generator/lib/src/generator.dart
@@ -1,5 +1,5 @@
 ///@nodoc
-import 'dart:async';
+import 'dart:async' show FutureOr;
 
 import 'package:analyzer/dart/constant/value.dart';
 import 'package:analyzer/dart/element/element.dart';
--- a/chopper_generator/lib/src/generator.dart
+++ b/chopper_generator/lib/src/generator.dart
@@ -1,4 +1,3 @@
-///@nodoc
 import 'dart:async' show FutureOr;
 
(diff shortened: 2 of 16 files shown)


chopper_generator/lib/src/generator.dart
```diff
@@ -1,59 +1,54 @@
...
 
-  bool _extendsChopperService(InterfaceType type) =>
+  static bool _extendsChopperService(InterfaceType type) =>
       _typeChecker(chopper.ChopperService).isExactlyType(type);
 
-  Field _buildDefinitionTypeMethod(String superType) => Field(
+  static Field _buildDefinitionTypeMethod(String superType) => Field(
         (method) => method
           ..annotations.add(refer('override'))
           ..name = 'definitionType'
```
Guldem, line 46: Isn't this also sort of a `Vars`?

chopper_generator/lib/src/generator.dart
```diff
@@ -466,100 +477,89 @@ class ChopperGenerator extends GeneratorForAnnotation<chopper.ChopperApi> {
...
+          if (hasParts) ...{
+            'parts': refer(Vars.parts.toString()),
+            'multipart': literalBool(true),
+          },
+          if (useQueries) 'parameters': refer(Vars.parameters.toString()),
+          if (useHeaders) 'headers': refer(Vars.headers.toString()),
+          if (useBrackets) 'useBrackets': literalBool(useBrackets),
+          if (includeNullQueryVars)
+            'includeNullQueryVars': literalBool(includeNullQueryVars),
+        },
```
Guldem, line 540: Nice! I like it!

Guldem (commented):
Nice this makes the generator more approachable! Hopefully it will also help possible transition into dart macro's/augmentations when its released.

chopper_generator/lib/src/generator.dart
techouse: Yeah, these methods could be further extracted into their own utility classes.

Guldem (approved):
LGTM! Besides my previous comments I don't see any blocking things. 

Only had to manual check if the output of the generated `test_service.chopper.dart` is the same. The CI currently doesn't verify/test if the generator create a different output and the dev didn't include the new file.

techouse:
> The CI currently doesn't verify/test if the generator create a different output and the dev didn't include the new file.

I might have a solution for that :)

Guldem:
Now I'm curious :D But I also have some ideas otherwise :)

techouse:
> Now I'm curious :D But I also have some ideas otherwise :)

https://pub.dev/packages/build_verify

Guldem:
> > Now I'm curious :D But I also have some ideas otherwise :)
> 
> https://pub.dev/packages/build_verify

Nice didn't know that one yet

techouse:
@JEuler can you take a look at this PR once you can? 🙃

JEuler:
Sure, I will. :)

chopper/test/ensure_build_test.dart
```diff
@@ -0,0 +1,16 @@
...
+  test(
+    'ensure_build',
+    () => expectBuildClean(
+      packageRelativeDirectory: 'chopper',
+      gitDiffPathArguments: [
+        'test/test_service.chopper.dart',
+      ],
+    ),
+  );
+}
```
Guldem, line 16: Is it intended to test it "twice". Since its also added in the `chopper_generator`? I would think it makes sense if the verify only takes place at one place. In this case the the `chopper_generator` since we want to be sure no unexpected changes happen.

JEuler (approved):

JEuler:
So, nothing bothers me at all. Maybe we can join buildOptionalPositionalParam and other one but no need in that. Thank you for the great contribution!

chopper/test/ensure_build_test.dart
techouse: Well, the thing is that both packages can have their own `test/test_service.dart` files which could be different. This one here in `chopper_generator` is just to test the generation of various methods.

I tried running a test from within `chopper_generator` referencing `../chopper/test/test_service.chopper.dart` directly, but the test wouldn't run correctly. And rather than using dodgy symlinks, I opted for a copy.  At the moment I think it's "good enough" 🙈 to copy the one from `chopper/tests/test_service.dart` to here. Some tests always beat no tests IMHO 🍡

techouse:
> Maybe we can join buildOptionalPositionalParam and other one but no need in that.

Yeah, I'll leave that for another day. I tried avoiding any logic changes here.

PR #451, opened 2023-07-07
authentication small bugfix

animousen4:
Original request used as first argument of send function, but should be modified (intercepted) request (req).


techouse:
Hey!

Nice spot!

techouse:
Confirmed. This is no bug.

PR #453, opened 2023-07-29
:sparkles: implement Dart 3 class modifiers

techouse:
- bump minimum Dart SDK version to 3.0.0
- bump dependencies to the latest (except `analyzer`)
- implement [Dart 3 class modifiers](https://dart.dev/language/class-modifiers)
- bump all package major versions:
  - chopper: 7.0.0
  - chopper_generator: 7.0.0
  - chopper_built_value: 2.0.0

### DIFF
--- a/chopper/lib/src/annotations.dart
+++ b/chopper/lib/src/annotations.dart
@@ -20,7 +20,7 @@ import 'package:meta/meta.dart';
 ///
 /// See [Method] to define an HTTP request
 @immutable
-class ChopperApi {
+final class ChopperApi {
   /// A part of a URL that every request defined inside a class annotated with [ChopperApi] will be prefixed with.
   final String baseUrl;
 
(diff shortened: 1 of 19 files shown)


techouse:
@JEuler I have added [class modifiers](https://dart.dev/language/class-modifiers) to make class inheritance more strict where I thought it necessary. Let me know your thoughts.

JEuler:
What do you think about sdk: ">=3.0.0 <4.0.0"? That is a big thing as I understand. Personally I don't have problems with that already, but maybe for someone that will be a problem. 

Any way, we should make it major then. :)

techouse:
> What do you think about sdk: ">=3.0.0 <4.0.0"?

Class modifiers are a Dart 3 thing. Dart 2 doesn't support them.

> Any way, we should make it major then. :)

Yep.

- chopper: 7.0.0
- chopper_generator: 7.0.0
- chopper_built_value: 2.0.0

JEuler:
Yeah, I know about the Dart 2-3. That's cool then, thank you!

JEuler (approved):

PR #480, opened 2023-08-13
:adhesive_bandage: export ChopperLogRecord to library surface

Maxr1998:
Forgot to do this is my last PR.

### DIFF
--- a/chopper/lib/chopper.dart
+++ b/chopper/lib/chopper.dart
@@ -10,6 +10,7 @@ export 'src/constants.dart';
 export 'src/extensions.dart';
 export 'src/interceptor.dart';
 export 'src/http_logging_interceptor.dart';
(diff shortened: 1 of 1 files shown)


techouse (approved):

techouse:
LGTM

As a side note, you could make `ChopperLogRecord` final, i.e.

```dart
final class ChopperLogRecord {
  const ChopperLogRecord(this.message, {this.request, this.response});

  final String message;
  final Request? request;
  final Response? response;

  @override
  String toString() => message;
}
```

as I do not believe that it should be extended nor implemented outside the library.

Maxr1998:
I think `final` is good. Opened #481 for that change since it's not directly related to this one.

JEuler (approved):

PR #493, opened 2023-08-29
Adding support for top level variables

martinale14:
I created this pr to give support to top level variables for the scenarios where we have multiple urls and that urls depend on ambient. for example we have our service A to make the main requests but we have service B to specific funcionality with diferent base url and maybe we want to change the service B to a dev version, so to change the url we have tu rebuild the chopper files because base url points to the literal value of the variable, with this change we don't need to rebuild in that scenerio because we can point the base url to a top level variable and that variable is going to be present in the generated file instead of the literal value

Imaging this case

```dart
const String enviromentUrl = Enviroments.temporal;

class Enviroments {
  static const String temporal = 'https://api.example.app';
}
```

```dart
@ChopperApi(baseUrl: enviromentUrl)
abstract class TeacherDatasource extends ChopperService {
  @Get(path: '/get_teachers/start/open')
  Future<Response<List<TeacherModel>>> getTeachers();

  static TeacherDatasource create([ChopperClient? client]) =>
      _$TeacherDatasource(client);
}
```

Before the change:

```dart
final class _$TeacherDatasource extends TeacherDatasource {
  _$TeacherDatasource([ChopperClient? client]) {
    if (client == null) return;
    this.client = client;
  }

  @override
  final definitionType = TeacherDatasource;

  @override
  Future<Response<List<TeacherModel>>> getTeachers() {
    final Uri $url =
        Uri.parse('https://api.example.app/get_teachers/start/open');
    final Request $request = Request(
      'GET',
      $url,
      client.baseUrl,
    );
    return client.send<List<TeacherModel>, TeacherModel>($request);
  }
}
```

After the change:

```dart
final class _$TeacherDatasource extends TeacherDatasource {
  _$TeacherDatasource([ChopperClient? client]) {
    if (client == null) return;
    this.client = client;
  }

  @override
  final definitionType = TeacherDatasource;

  @override
  Future<Response<List<TeacherModel>>> getTeachers() {
    final Uri $url = Uri.parse('${enviromentUrl}/get_teachers/start/open');
    final Request $request = Request(
      'GET',
      $url,
      client.baseUrl,
    );
    return client.send<List<TeacherModel>, TeacherModel>($request);
  }
}
```

### DIFF
--- a/chopper/lib/src/annotations.dart
+++ b/chopper/lib/src/annotations.dart
@@ -22,6 +22,8 @@ import 'package:meta/meta.dart';
 @immutable
 final class ChopperApi {
   /// A part of a URL that every request defined inside a class annotated with [ChopperApi] will be prefixed with.
+  ///
+  /// The `baseUrl` can be a top level constant string variable.
   final String baseUrl;
 
   const ChopperApi({
--- a/chopper/test/base_test.dart
+++ b/chopper/test/base_test.dart
@@ -9,8 +9,10 @@ import 'package:http/testing.dart';
 import 'package:test/test.dart';
 
 import 'test_service.dart';
+import 'test_service_variable.dart';
 
 final baseUrl = Uri.parse('http://localhost:8000');
+const String testEnv = 'https://localhost:4000';
 
 void main() {
   ChopperClient buildClient([
@@ -22,6 +24,7 @@ void main() {
         services: [
           // the generated service
           HttpTestService.create(),
+          HttpTestServiceVariable.create(),
         ],
         client: httpClient,
         errorConverter: errorConverter,
@@ -35,9 +38,12 @@ void main() {
 
       final chopper = buildClient(httpClient);
       final service = chopper.getService<HttpTestService>();
+      final serviceVariable = chopper.getService<HttpTestServiceVariable>();
 
       expect(service, isNotNull);
+      expect(serviceVariable, isNotNull);
       expect(service, isA<HttpTestService>());
+      expect(serviceVariable, isA<HttpTestServiceVariable>());
     });
 
     test('get service errors', () async {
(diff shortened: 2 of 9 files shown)


techouse:
Can you please elaborate on the change?

martinale14:
> Can you please elaborate on the change?

Thanks for the comment, I already edited the pr's comment

techouse:
> for example we have our service A to make the main requests but we have service B to specific funcionality with diferent base url and maybe we want to change the service B to a dev version, so to change the url we have tu rebuild the chopper files because base url points to the literal value of the variable

Maybe I don't fully understand, but to me this solution sounds like a hack and I'm not sure that it's the responsibility of Chopper to give the user the option of having multiple environments.

Why not simply abstract all these `baseUrl`s into multiple environment files and load those based on the desired scenario?

Sorry if I'm being too nosy, but can you maybe provide a few examples?

martinale14:
> > for example we have our service A to make the main requests but we have service B to specific funcionality with diferent base url and maybe we want to change the service B to a dev version, so to change the url we have tu rebuild the chopper files because base url points to the literal value of the variable
> 
> Maybe I don't fully understand, but to me this solution sounds like a hack and I'm not sure that it's the responsibility of Chopper to give the user the option of having multiple environments.
> 
> Why not simply abstract all these `baseUrl`s into multiple environment files and load those based on the desired scenario?
> 
> Sorry if I'm being too nosy, but can you maybe provide a few examples?

i updated the comment again with an example

techouse:
> i updated the comment again with an example, no problem bro

I've cleaned it up for you :)

I see what you mean. I'd usually solve this using a `ChopperClient` instance, however, this seems like a viable solution as well.

Right, can you add some tests for this, then I think we can merge it :)

martinale14:
> > i updated the comment again with an example, no problem bro
> 
> I've cleaned it up for you :)
> 
> I see what you mean. I'd usually solve this using a `ChopperClient` instance, however, this seems like a viable solution as well.
> 
> Right, can you add some tests for this, then I think we can merge it :)

ready man i upload some tests

techouse (approved):
LGTM

@JEuler your thoughts?

Guldem:
I don't think this addition doesn't do any harm. It doesn't directly change the way how some one would use chopper. Maybe a bit more documentation could be added in the doc files. So its not really a "hidden" feature. 

Although I think I would have solved this by rewriting baseUrl in a Interceptor. If I had to change thinks runtime.

Issue #519, opened 2023-10-12
Incorrect url generation when using new baseurl


martinale14:
## Steps to Reproduce

1. Actual builder if you provide this input

```dart
@ChopperApi(baseUrl: 'https://localhost:4000/test')
abstract class HttpTestServiceBaseUrl extends ChopperService {
  static HttpTestServiceBaseUrl create([ChopperClient? client]) =>
      _$HttpTestServiceBaseUrl(client);

  @Get(path: '')
  Future<Response> getAll();
}
```
2. Produces this output
```dart
final class _$HttpTestServiceBaseUrl extends HttpTestServiceBaseUrl {
  _$HttpTestServiceBaseUrl([ChopperClient? client]) {
    if (client == null) return;
    this.client = client;
  }

  @override
  final definitionType = HttpTestServiceBaseUrl;

  @override
  Future<Response<dynamic>> getAll() {
    final Uri $url = Uri.parse('https:/localhost:4000/test');
    final Request $request = Request(
      'GET',
      $url,
      client.baseUrl,
    );
    return client.send<dynamic, dynamic>($request);
  }
}
```
3. As you notice the final Uri is generating with only one slash ```https:/...```
4. I am currently working on a solution

PR #520, opened 2023-10-12
Fix incorrect url generation when using new baseurl

martinale14:
I rewrite the replaceallMethod that removes double ```//``` making it follow a regular expresion that ignores ```http://``` and ```https://``` on the replacement

---

Addresses #519

### DIFF
--- a/chopper/CHANGELOG.md
+++ b/chopper/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 7.0.9
 
-- Fix ChopperGenerator building ```https://...``` with one slash ([#519](https://github.com/lejard-h/chopper/pull/519))
+- Fix ChopperGenerator building ```https://...``` with one slash ([#520](https://github.com/lejard-h/chopper/pull/520))
 
 ## 7.0.8
 
--- a/chopper/CHANGELOG.md
+++ b/chopper/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 7.0.9
+
+- Fix ChopperGenerator building ```https://...``` with one slash ([#519](https://github.com/lejard-h/chopper/pull/519))
+
 ## 7.0.8
 
 - Encode DateTime query parameters in ISO8601 format ([#516](https://github.com/lejard-h/chopper/pull/516))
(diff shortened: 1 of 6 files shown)


chopper/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 7.0.9
```
techouse, line 3: Can you please remove this entry? I'll add it later once I make the actual release.

techouse (changes requested):

chopper_generator/lib/src/generator.dart
```diff
@@ -550,7 +550,12 @@ final class ChopperGenerator
       }
     }
 
-    return _generateUri('$finalBaseUrl$path'.replaceAll('//', '/'));
+    return _generateUri(
+      '$finalBaseUrl$path'.replaceAll(
+        RegExp(r'(?<!http:|https:)\/\/'),
```
techouse, line 555: I just put this REGEX into regexr.com and have seen that it will match:

- https:`//`test.com/foo/bar/baz
- http://test.com/foo/bar/baz
- https:`//`test.com`//`foo`//`bar`//`baz
- http://test.com`//`foo`//`bar`//`baz

Is this expected?

martinale14: look i did the same and works as expected

image

maybe is your browser, could you please tell me what browser are you using?

chopper/CHANGELOG.md
martinale14: Remove all the 7.0.9 section or only the line 3

chopper_generator/lib/src/generator.dart
techouse: The latest Safari.

techouse: I still find this regex too cryptic for my taste. Have you tried this approach

1. extract the protocol using `Uri.scheme`
2. find how long the scheme and intial `//` is, i.e. 6 or 7 chars
3. use `substring` to replace all `//` with `/` in the remainder od the string without the protocol and initial `//`

martinale14: ok bro i will implement that solution

martinale14: ready

techouse:
I'll make a release later in the evening.

techouse:
@martinale14 ^^ released

PR #534, opened 2023-11-15
:lipstick: cosmetic generator refactor

techouse:
I scanned through the generator again and refactored some bits and pieces.

@JEuler let me know your thoughts on this.

### DIFF
--- a/chopper/example/definition.chopper.dart
+++ b/chopper/example/definition.chopper.dart
@@ -6,6 +6,7 @@ part of 'definition.dart';
 // ChopperGenerator
 // **************************************************************************
 
+// coverage:ignore-file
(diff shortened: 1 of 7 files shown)


JEuler:
```
  old: refer('${Vars.client}.${Vars.baseUrl}'),
  new: refer(Vars.client.toString()).property(Vars.baseUrl.toString()),
```

It becomes longer without any like, typing benefits, I think. Is that necessary after the refactoring?

techouse:
The intention here is to swap direct String manipulation with something "safer" 😅

JEuler:
Got it. :)

JEuler (approved):

Issue #429, opened 2023-05-24
[Feature request] Custom error generic type please?


urusai88:
With chopper we should manually convert Response.error object. Also we must catch the critical exceptions.
Can we have please a mechanism to catch critical errors and convert Response.error to generic type object?
Sorry for my bad english, i hope you understood me.

techouse:
@urusai88 can you please provide an example?

urusai88:
> @urusai88 can you please provide an example?

```dart
class MyCustomError {}

final chopper = ChopperClient<MyCustomError>(
    
);

final service = chopper.getService<...>();
final resp = await service.get();
if(!resp.isSuccessful) {
    // resp.error is MyCustomError? here
}
```

Guldem:
Not sure I fully understand. But its possible to use a `errorConverter` which handles error responses (statusCode < 200 || statusCode >= 300). The errorConverter can return a custom error object. Only the client/service doesn't accept a generics for error types.  But its possible to cast the error from a response to the then desired custom object.

Something like:
```dart
class MyErrorConverter implements ErrorConverter  {

  @override
  FutureOr<Response> convertError<BodyType, InnerType>(Response response){
    // do something with error
    return MyCustomErrorClass();
  }
}

final chopper = ChopperClient(errorConverter: MyErrorConverter(), ...);

final service = chopper.getService<...>();
final resp = await service.get();
if(!resp.isSuccessful) {
  final myError =  resp.error as MyCustomErrorClass;
}
```

urusai88:
@Guldem 
It is easy solution, but not type safe and forces casting every time

Guldem:
The hard think with adding a generic type for errors is that generic need to be defined. So even if you don't care about custom errors you need to pass a type somewhere. Another solution might be to add a function/getter to response that does the logic (type checking/casting for you).

Something like(probably needs better naming): 
```dart
T? get errorOrNull<T>  {
  if(error != null && error is T) {
    return error as T;
  } else {
    return null;
  }
}
```

Then it would be something like:
```dart
final service = chopper.getService<...>();
final resp = await service.get();
if(!resp.isSuccessful && resp.errorOrNull<MyCustomError> != null) {
    // do something with your error.
}
```

You could also add this as an extension on response if you want to test it.

Guldem:
@urusai88 The suggestion I made is now also added to chopper itself. Does this solve your request? :)

urusai88:
@Guldem yes, thanks you

PR #543, opened 2024-01-02
Helper function for fetching errors of specific type

Guldem:
Added simple helper function for fetching error of a specific type. This could be helpful when a error converter is added to convert error into a custom error object. This reduces the boilerplate needed to get the custom error from the response.

See #429

### DIFF
--- a/chopper/lib/src/response.dart
+++ b/chopper/lib/src/response.dart
@@ -67,6 +67,15 @@ base class Response<BodyType> with EquatableMixin {
   String get bodyString =>
       base is http.Response ? (base as http.Response).body : '';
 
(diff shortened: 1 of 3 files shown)


techouse (approved):

techouse:
LGTM. You could add some more explicit tests, though.  😄

Guldem:
> LGTM. You could add some more explicit tests, though. 😄

Not sure what you mean? I find them pretty explicit haha 😅

techouse:
>  Could improve to test names though

That. It just tests for `int` at the moment. Make a `FooError` or something. :)

PR #545, opened 2024-01-03
:sparkles: Feature: Omit Response in service

Guldem:
Added the possibility to omit Response when creating a service to reduce some boilerplate if the Response object is not really needed.

This give the developer the choice to make use of the Response and its meta data is needed. This might not always be case because a error converter deals with the error logic. 

**TODO**:
Update documentation for this feature

Previously a `ChopperService` was created like this (still valid):
```dart
@ChopperApi(baseUrl: '/test')
abstract class MyChopperService extends ChopperService {
  static MyChopperService create([ChopperClient? client]) =>
      _$MyChopperService(client);

  @Get(path: 'get/{id}')
  Future<Response<String>> getSomething(
    @Path() String id, {
    @Header('test') required String dynamicHeader,
  });
}
```

Which creates:
```dart
final class _$MyChopperService extends MyChopperService {
  _$MyChopperService([ChopperClient? client]) {
    if (client == null) return;
    this.client = client;
  }

  @override
  final Type definitionType = MyChopperService;

  @override
  Future<Response<String>> getSomething(
    String id, {
    required String dynamicHeader,
  }) {
    final Uri $url = Uri.parse('/test/get/${id}');
    final Map<String, String> $headers = {
      'test': dynamicHeader,
    };
    final Request $request = Request(
      'GET',
      $url,
      client.baseUrl,
      headers: $headers,
    );
    return client.send<String, String>($request);
  }
```

Now its also possible to specify a service like this:
```dart
@ChopperApi(baseUrl: '/test')
abstract class MyChopperService extends ChopperService {
  static MyChopperService create([ChopperClient? client]) =>
      _$MyChopperService(client);

  @Get(path: 'get/{id}')
  Future<String> getSomething(
    @Path() String id, {
    @Header('test') required String dynamicHeader,
  });
}
```

Which creates this:
```dart
final class _$MyChopperService extends MyChopperService {
  _$MyChopperService([ChopperClient? client]) {
    if (client == null) return;
    this.client = client;
  }

  @override
  final Type definitionType = MyChopperService;

  @override
  Future<String> getSomething(
    String id, {
    required String dynamicHeader,
  }) async {
    final Uri $url = Uri.parse('/test/get/${id}');
    final Map<String, String> $headers = {
      'test': dynamicHeader,
    };
    final Request $request = Request(
      'GET',
      $url,
      client.baseUrl,
      headers: $headers,
    );
    final Response $response = await client.send<String, String>($request);
    return $response.bodyOrThrow;
  }
```

### DIFF
--- a/chopper/lib/src/response.dart
+++ b/chopper/lib/src/response.dart
@@ -1,3 +1,4 @@
+import 'dart:io';
 import 'dart:typed_data';
 
 import 'package:equatable/equatable.dart' show EquatableMixin;
@@ -73,6 +74,20 @@ base class Response<BodyType> with EquatableMixin {
       return error as ErrorType;
     } else {
       return null;
+    }
+  }
+
+  /// Returns the response body if [Response] [isSuccessful] and [body] is not null.
+  /// Otherwise it throws an [HttpException] with the response status code and error object.
+  /// If the error object is an [Exception], it will be thrown instead.
+  BodyType get bodyOrThrow {
+    if (isSuccessful && body != null) {
+      return body!;
+    } else {
+      if (error is Exception) {
+        throw error!;
+      }
+      throw HttpException('Could not fetch response $statusCode: $error');
     }
   }
 
--- a/chopper/lib/src/response.dart
+++ b/chopper/lib/src/response.dart
@@ -1,6 +1,7 @@
 import 'dart:io';
 import 'dart:typed_data';
 
+import 'package:chopper/src/chopper_http_exception.dart';
 import 'package:equatable/equatable.dart' show EquatableMixin;
 import 'package:http/http.dart' as http;
 import 'package:meta/meta.dart';
@@ -87,7 +88,7 @@ base class Response<BodyType> with EquatableMixin {
       if (error is Exception) {
         throw error!;
       }
-      throw HttpException('Could not fetch response $statusCode: $error');
+      throw ChopperHttpException(this);
     }
   }
 
--- a/chopper/lib/src/response.dart
+++ b/chopper/lib/src/response.dart
@@ -1,4 +1,3 @@
-import 'dart:io';
 import 'dart:typed_data';
 
 import 'package:chopper/src/chopper_http_exception.dart';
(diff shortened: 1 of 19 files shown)


chopper/lib/src/response.dart
```diff
@@ -1,3 +1,4 @@
+import 'dart:io';
```
techouse, line 1: This excludes browser based Chopper apps since `dart:io` is only available for:

- Servers
- Command-line scripts
- Flutter mobile apps
- Flutter desktop apps

https://api.dart.dev/stable/3.2.4/dart-io/dart-io-library.html

techouse:
Nice addition!

What's the plan for APIs that don't return anything, i.e. a `DELETE` request that only returns an HTTP status code `204`?

```dart
@Delete(path: 'get/{id}')
Future<void> getSomething(@Path() String id);
```

Like this ⬆️?

chopper/lib/src/response.dart
Guldem: Forgot it.. added custom exception instead

Guldem:
> Nice addition!
> 
> What's the plan for APIs that don't return anything, i.e. a `DELETE` request that only returns an HTTP status code `204`?
> 
> ```dart
> @Delete(path: 'items/{id}')
> Future<void> deleteItem(@Path() String id);
> ```
> 
> Like this ⬆️?

When the service has `Future<void>` this will also be generated in the service implementation. I only wasn't able to find what the actual body becomes when a `NoContent`/`204` occurs. 

Added a extra tests for `bodyOrThrow` to test it. But this fail if the body would be set to `null` then it would throw the exception. But looking at this [line](https://github.com/lejard-h/chopper/blob/4ba5a83654388a33f70cacf7d8619880cd573523/chopper/lib/src/base.dart#L309) it would expect atleast a empty `String`.

Not complete sure though.

chopper/lib/src/response.dart
techouse: The import still exists though. 😅

Guldem: Seriously 🤦  OMG...

techouse (approved):
LGTM

techouse:
Do you plan to update the docs in this PR?

Guldem:
Was thinking on doing separate separate PR for updating docs and include some more information on creating chopper services.

PR #567, opened 2024-01-25
:sparkles: Add Target annotations

Guldem:
Added `Target` annotations to chopper annotations to provide analysers and tooling with more information about the annotation placements.


techouse:
I'm not familiar with this feature. What does it do exactly?

Guldem:
The dart analyzer can use the provided context to determine if the annotation placement is correct. For example placing the `@Body()` annotation on a class instead of a method parameter will result in a lint [invalid_annotation_target](https://dart.dev/tools/diagnostic-messages#invalid_annotation_target) warning.

image
image

techouse (approved):

techouse:
Cool! LGTM!

PR #570, opened 2024-01-31
:bug: Export ChopperHttpException

Guldem:
`chopper_http_exception.dart` wasn't included in library export. Which makes it hard to catch in try catch statements because importing is harder.


techouse (approved):
LGTM

Issue #577, opened 2024-02-29
[Feature Request] Add @FormUrlEncoded


lwj1994:
Consider deprecating FormUrlConverter and use the `@FormUrlEncoded` annotation to move FormUrlConverter processing to code generation.

```
  @Post(path: 'formUrlEncoded')
  @FormUrlEncoded()
  Future<Response> postFormUrlEncode(
    @Field('a') String a,
    @FieldMap() Map<String, String> b,
  );
```

PR #579, opened 2024-03-01
Support @FormUrlEncoded

lwj1994:
fix https://github.com/lejard-h/chopper/issues/577

* Support @FormUrlEncoded，
* auto add content-type application/x-www-form-urlencoded 
* set ket.toString and value.toString() for @Body / @Field / @FieldMap data,  if data is `Map` and  `Map` is not `Map<String, String>`

### DIFF
--- a/chopper/lib/src/annotations.dart
+++ b/chopper/lib/src/annotations.dart
@@ -536,13 +536,13 @@ final class PartFileMap {
 /// {@template FormUrlEncoded}
 ///
 ///
-//  Denotes that the request body will use form URL encoding. Fields should be declared as parameters
-//  and annotated with [Field]/[FieldMap].
-//
-//  Requests made with this annotation will have application/x-www-form-urlencoded MIME
-//  type. Field names and values will be UTF-8 encoded before being URI-encoded in accordance to <a
-//  href="https://datatracker.ietf.org/doc/html/rfc3986">RFC-3986</a>.
-//
+/// Denotes that the request body will use form URL encoding. Fields should be declared as parameters
+/// and annotated with [Field]/[FieldMap].
+///
+/// Requests made with this annotation will have application/x-www-form-urlencoded MIME
+/// type. Field names and values will be UTF-8 encoded before being URI-encoded in accordance to <a
+/// href="https://datatracker.ietf.org/doc/html/rfc3986">RFC-3986</a>.
+///
 ///
 /// ```dart
 /// @Post(path: '/something')
--- a/chopper/lib/src/annotations.dart
+++ b/chopper/lib/src/annotations.dart
@@ -433,7 +433,7 @@ final class Field {
 ///
 /// ```dart
 /// @Post(path: '/something')
-/// Future<Response> fetch(@FieldMap List<Map<String, dynamic>> query);
+/// Future<Response> fetch(@FieldMap Map<String, dynamic> query);
 /// ```
 /// {@endtemplate}
 @immutable
@@ -533,6 +533,30 @@ final class PartFileMap {
   const PartFileMap();
 }
 
+/// {@template FormUrlEncoded}
+///
+///
+//  Denotes that the request body will use form URL encoding. Fields should be declared as parameters
+//  and annotated with [Field]/[FieldMap].
+//
+//  Requests made with this annotation will have application/x-www-form-urlencoded MIME
+//  type. Field names and values will be UTF-8 encoded before being URI-encoded in accordance to <a
+//  href="https://datatracker.ietf.org/doc/html/rfc3986">RFC-3986</a>.
+//
+///
+/// ```dart
+/// @Post(path: '/something')
+/// @FormUrlEncoded
+/// Future<Response> fetch(@Field("param") String? param);
+/// ```
+/// {@endtemplate}
+@immutable
+@Target({TargetKind.method})
+final class FormUrlEncoded {
+  /// {@macro FormUrlEncoded}
+  const FormUrlEncoded();
+}
+
 /// {@macro ChopperApi}
 const chopperApi = ChopperApi();
 
@@ -595,3 +619,6 @@ const partFile = PartFile();
 
 /// {@macro PartFileMap}
 const partFileMap = PartFileMap();
+
+/// {@macro FormUrlEncoded}
+const formUrlEncoded = FormUrlEncoded();
--- a/chopper_generator/lib/src/generator.dart
+++ b/chopper_generator/lib/src/generator.dart
@@ -150,6 +150,7 @@ final class ChopperGenerator
   ) {
     final ConstantReader? method = _getMethodAnnotation(m);
     final bool multipart = _hasAnnotation(m, chopper.Multipart);
+    final bool formUrlEncoded = _hasAnnotation(m, chopper.FormUrlEncoded);
     final ConstantReader? factoryConverter = _getFactoryConverterAnnotation(m);
 
     final Map<String, ConstantReader> body = _getAnnotation(m, chopper.Body);
@@ -172,7 +173,7 @@ final class ChopperGenerator
     final Map<String, ConstantReader> fileFieldMap =
         _getAnnotation(m, chopper.PartFileMap);
 
-    final Code? headers = _generateHeaders(m, method!);
+    final Code? headers = _generateHeaders(m, method!, formUrlEncoded);
     final Expression url = _generateUrl(
       method,
       paths,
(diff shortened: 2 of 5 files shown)


chopper_generator/lib/src/generator.dart
```diff
@@ -466,10 +488,41 @@ final class ChopperGenerator
...
+      final mapTypeChecker = TypeChecker.fromRuntime(Map);
+      if (!mapTypeChecker.isExactlyType(type)) {
+        return;
+      }
+    }
+    final mapType = type?.toString() ?? 'Map<String, dynamic>';
+    final idNullableValueType =
+        mapType.endsWith('dynamic>') || mapType.endsWith('?>');
+    final isStringValueType = mapType.endsWith('String>');
+    if (!idNullableValueType && isStringValueType) {
```
Guldem, line 503: Not sure if this is the nicest way of checking the types. It's possible to check inner type as far I know. See [Response type checker](https://github.com/lejard-h/chopper/blob/2249102880d84f21854475bd641f4d5ce336f0c5/chopper_generator/lib/src/generator.dart#L602).

 But doesn't `mapType.endsWith('String>')` also mean its non-nullable. So the if statement only needs to check if its a `isStringValueType`.

chopper_generator/lib/src/generator.dart
```diff
@@ -466,10 +488,41 @@ final class ChopperGenerator
...
+      return;
+    }
+    final cleanBody = '${refName}Clean';
+    blocks.add(Code('final $mapType $cleanBody = {};'));
+    blocks.add(Code('$refName.forEach((key, value) {'));
```
Guldem, line 508: `forEach` is not always preferred. See [lint rule](https://dart.dev/tools/linter-rules/avoid_function_literals_in_foreach_calls).

Guldem (changes requested):

chopper/lib/src/request.dart
```diff
@@ -158,6 +158,13 @@ base class Request extends http.BaseRequest with EquatableMixin {
         request
           ..headers.addAll(headers)
           ..bodyFields = body;
+      } else if (body is Map<dynamic, dynamic> &&
+          headers['content-type'] == 'application/x-www-form-urlencoded') {
+        final newBody = (body as Map<dynamic, dynamic>).map<String, String>(
+            (key, value) => MapEntry(key.toString(), value.toString()));
+        request
+          ..headers.addAll(headers)
+          ..bodyFields = newBody;
```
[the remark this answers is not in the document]
lwj1994: set the cleanBody type to Map<String, String> for formurlencode

chopper_generator/lib/src/generator.dart
```diff
@@ -287,6 +288,11 @@ final class ChopperGenerator
       }
 
       final bool hasQuery = hasQueryMap || queries.isNotEmpty;
+      DartType? bodyType;
+
+      if (hasQuery) {
+        _cleanMapBody(blocks, Vars.parameters.toString(), type: bodyType);
+      }
```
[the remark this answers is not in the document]
lwj1994: removed

chopper_generator/lib/src/generator.dart
```diff
@@ -466,6 +485,43 @@ final class ChopperGenerator
...
+    if (!handleToString && !handleNullableValue) {
+      return null;
+    }
+
+    final bodyType = type?.toString() ?? 'Map<String, dynamic>';
+    final cleanType = valueToString ? 'Map<String, String>' : bodyType;
+    final cleanSymbol = '${refName}Clean';
+    blocks.add(Code('final $cleanType $cleanSymbol = {};'));
+    blocks.add(Code('for (var key in $refName.keys) {'));
+    if (handleNullableValue) {
```
Guldem, line 513: Not sure if nullable need to be checked. When looking at the "old" `FormUrlEncodedConverter` it doesn't care about null values.

Guldem (changes requested):

chopper_generator/lib/src/generator.dart
lwj1994: > Not sure if nullable need to be checked. When looking at the "old" `FormUrlEncodedConverter` it doesn't care about null values.

null values are meaningless in most scenarios, maybe add an optional bool switch？

chopper_generator/lib/src/generator.dart
```diff
@@ -298,6 +300,11 @@ final class ChopperGenerator
       bool hasBody = body.isNotEmpty || fields.isNotEmpty;
       if (hasBody) {
         if (body.isNotEmpty) {
+          for (var element in m.parameters) {
+            if (element.name == refer(body.keys.first).symbol) {
+              bodyType = element.type;
+            }
+          }
```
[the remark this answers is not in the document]
lwj1994: > I think this should be in the `else` part?
it  find the Class type of the @Body annotation.  After the test, the result was correct

chopper_generator/lib/src/generator.dart
```diff
@@ -466,6 +485,43 @@ final class ChopperGenerator
     });
   }
 
+  static Reference? _generateCleanMapBody(
```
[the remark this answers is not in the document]
lwj1994: rewrited

chopper_generator/lib/src/generator.dart
```diff
@@ -298,33 +299,44 @@ final class ChopperGenerator
...
           );
         }
       }
 
       final bool hasFieldMap = fieldMap.isNotEmpty;
       if (hasFieldMap) {
+        final DartType fieldMapType = m.parameters
+            .firstWhere(
+                (p) => _typeChecker(chopper.FieldMap).hasAnnotationOf(p))
+            .type;
```
Guldem, line 327: Doesn't the `fieldMap` variable also exposed the `DartType`? Via `
fieldMap.values.first.typeValue;`?

chopper_generator/lib/src/generator.dart
```diff
@@ -298,33 +299,44 @@ final class ChopperGenerator
...
+        final map = (formUrlEncoded && !_isMapStringString(fieldMapType))
+            ? _generateMapToStringExpression(refer(fieldMap.keys.first))
+            : refer(fieldMap.keys.first);
```
Guldem, line 330: Question: What if you add multiple `@FieldMap()` annotation in one call? Is that something that's possible? Or should we throw a warning?

chopper_generator/lib/src/generator.dart
```diff
@@ -664,17 +722,20 @@ final class ChopperGenerator
       );
 
   static Expression _generateMap(
-    Map<ParameterElement, ConstantReader> queries,
-  ) =>
+    Map<ParameterElement, ConstantReader> queries, {
+    bool isStringValue = false,
```
Guldem, line 726: Nit: Maybe change name to `enableToString` or something more descriptive?

chopper/lib/src/annotations.dart
```diff
@@ -433,7 +433,7 @@ final class Field {
 ///
 /// ```dart
 /// @Post(path: '/something')
-/// Future<Response> fetch(@FieldMap List<Map<String, dynamic>> query);
+/// Future<Response> fetch(@FieldMap Map<String, dynamic> query);
```
Guldem, line 436: Was this even supported?

Guldem (changes requested):
Nice additions! 👍  Looking good only got a few small questions

chopper_generator/lib/src/generator.dart
lwj1994: > Doesn't the `fieldMap` variable also exposed the `DartType`? Via ` fieldMap.values.first.typeValue;`?

i tried it but get error: `FormatException: Not an instance of Type.`

```dart
        try{
          _logger.info("fieldMapType = ${fieldMap.values.first.typeValue.toString()}");
        } catch (e) {
          _logger.info("error fieldMapType  \n${e}");
        }
```

run and output :
```dart
[INFO] Chopper Generator:error fieldMapType
FormatException: Not an instance of Type.
```

chopper/lib/src/annotations.dart
lwj1994: There is no support for `List<Map<String, dynamic>>` , I think it was a mistake.

chopper_generator/lib/src/generator.dart
lwj1994: in line 515. ConstantReader receive the  Annotation,  it is not a DartType，so it will crash when call `typeValue`. 
``` dart
    for (final ParameterElement p in method.parameters) {
      final DartObject? a = _typeChecker(type).firstAnnotationOf(p);
      if (annotation != null && a != null) {
        throw Exception(
          'Too many $type annotation for \'${method.displayName}\'',
        );
      } else if (annotation == null && a != null) {
        annotation = a;
        name = p.displayName;
      }
    }

    return annotation == null ? {} : {name: ConstantReader(annotation)};
  }
```

Guldem: Ok was just wondering. Thanks for looking into it!

Guldem (approved):
LGTM! Would be nice if we could also update docs. But that can also be a different MR.

lwj1994:
> LGTM! Would be nice if we could also update docs. But that can also be a different MR.

👌🏻. Leave it to me

Guldem (approved):
Thanks for adding docs 👍

techouse (changes requested):
Thanks for this. 😇

I added a few comments.

chopper_generator/test/test_service.dart
```diff
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:collection';
```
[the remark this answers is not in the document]
techouse: Nevermind. Wrong collection LOL

techouse (approved):
LGTM

techouse:
Thank you for your contribution @lwj1994

PR #547, opened 2024-01-05
:sparkles: Restructure interceptors

Guldem:
# Restructuring interceptors
This PR/Draft contains the current progress/experimenting on restructuring interceptors. See #544 for more information on the proposal. Any feedback/help is welcome!

## Description
Refactor interceptors into a chain of interceptors. This will reduce the amount of different type interceptor interfaces to 1. A `Interceptor` will be responsible for both the `Request` and the `Response` if needed.

Specifying a interceptor would look like this:
```dart
class MyInterceptor implements Interceptor {
 @override
 FutureOr<Response<BodyType>> intercept<BodyType>(Chain<BodyType> chain) async {
    final request = chain.request;
    final updatedRequest = // Do something with intercepted request here

    final response = await chain.proceed(updatedRequest);
    final updatedResponse =  // Do something with intercepted response.

    return updatedResponse;
  }
}
```

All `Interceptor` are chained so the `Request` will first go down the chain of interceptors and in the last step the network call is make. The `Response` will then go back through the chain of interceptors.

Besides user created interceptors some of the internal logic have been refactor into `InternalInterceptor`. 

InternalInterceptors:
- RequestConverterInterceptor: Wrapper interceptor that wraps request converters
- ResponseConverterInterceptor: Wrapper interceptor that wraps response converters
- HttpCallInterceptor: Interceptor that makes the actual http call
- AuthenticatorInterceptor: Wrapper interceptor that wraps the authenticator

## Progress
- [x] Setup the actual chain
- [x] Include converters in the chain
- [x] Include authenticator in the chain
- [x] Refactor `HeadersInterceptor`, `CurlInterceptor` and `HttpLoggingInterceptor` to use new interface.
- [x] Update HttpLoggingInterceptor to track call time.
- [x] Check/Fix exception handling
- [x] Update existing tests
- [x] Add new tests
- [x] Add dart doc to code
- [x] Add documentation

## Breaking changes
- RequestInterceptor is removed
- Function RequestInterceptors are removed
- ResponseInterceptor is removed
- Function ResponseInterceptors are removed

PR #586, opened 2024-03-18
Support @Tag

lwj1994:
Dynamic tag can be added to the request annotated with the `@Tag`
annotation. Default values are supported.

```dart
Future<Response> requestWithTag(@Tag() String t1, @Tag("ta") int t2);
```

get tag via `request.tags`:
```dart
final tags = request.tag;
final String t1 = tags["t1"] as String;
final int t2 = tags["t2"] as int;
```

If you prefer to pass a `Map` of tag, you can do so with the `@TagMap` annotation.

```dart
Future<Response> requestWithTag(@TagMap() Map<String, dynamic> tags);
```


techouse:
Hey,

Thanx for the PR, however, I'm not sure I understand what this `@Tag` annotation. Can you add some concrete examples and/or tests please?

lwj1994:
> Hey,
> 
> Thanx for the PR, however, I'm not sure I understand what this `@Tag` annotation should be used for.
> 
> Can you add some concrete examples and/or tests please?

ok.  i will add  it later.


Let me start with a small example .

some path need check token to trigger refresh token when token is out of date. 

we can determine the path in interceptor:
```dart
class  TokenInterceptor extends RequestInterceptor{
   request( ) {
      if (path.startWith("/xxx/xx")  
        || path.startWith("/xxx/xx")
        || path.startWith("/xxx/xx")
        || path.startWith("/xxx/xx")
       // ...... more
){
           // wait to refresh token 
          await ensureTokenIsValid();
          request.headers.add["Auth":  token];
      }
   }
}
```

If those token-dependent requests are tagged, then no judgment is required。

```dart
if (request.tags["needToken"] != null){
           // wait to refresh token 
          await ensureTokenIsValid();
          request.headers.add["Auth":  token];
      }
```

techouse:
Why don't you simply catch an expired token, refresh it and repeat the call once more?

```
```

lwj1994:
> Why don't you simply catch an expired token, refresh it and repeat the call once more?
> 
> ```
> (request) --> (error 401: token expired) --> (auth interceptor refresh token) --> (repeat request) --> (ok 200)
> ```

thx, it looks more simple，but the `BodyType` will lose type, just as `dynamic`,
```dart
class AuthInterceptor implements chopper.ResponseInterceptor {

  @override
  FutureOr<chopper.Response<dynamic>> onResponse(chopper.Response<dynamic> response) async{
    if(response.statusCode == 401){
      await refreshToken();
      // resend
      response = await chopperClient.send<BodyType>(response.base.request as chopper.Request);
    }
    return response;
  }
}
```

techouse:
> i noticed that 👍🏻 after #547 interceptor can get type.

Correct! Can you please test #547 and let @Guldem know if everything works for you. It's a bigger/breaking PR so we'd be grateful for any real-world tests. 😇

Guldem:
It's nice to add meta information to a request but I personally never needed something like this. Like @techouse says in most cases it can be solved with a interceptor. But on the other side it can't really hurt 😄

requests.md
```diff
@@ -276,4 +279,26 @@ Future<Response<MyClass>> fetch();
 Future<MyClass> fetch();
 ```
 
-> Note: Chopper doesn't convert response bodies by itself to dart object. You need to use a [Converter](converters/converters.md) for that. 
\ No newline at end of file
+> Note: Chopper doesn't convert response bodies by itself to dart object. You need to use a [Converter](converters/converters.md) for that.
+
+## Add tag
+Dynamic tag can be added to the request annotated with the `@Tag`
+annotation. Default values are supported.
```
Guldem, line 286: dynamic includes all types.

requests.md
```diff
@@ -276,4 +279,26 @@ Future<Response<MyClass>> fetch();
...
+final String t1 = tags["t1"] as String;
+final int t2 = tags["t2"] as int;
+```
+
+
+If you prefer to pass a `Map` of tag, you can do so with the `@TagMap` annotation.
+
+```dart
+Future<Response> requestWithTag(@TagMap() Map<String, dynamic> tags);
+```
```
Guldem, line 304: Maybe also describe how to access tags. And problem/use-case these tags have

lwj1994:
hi guys, i refactored tag,  make tag as a Object instead of Map<String, dynamic>.  and  i add my showcase in  `chopper/example/tag.dart`.  Although not very useful, some cases are used as a means of tracking requests.

## Add tag
`@Tag` parameter annotation for setting tag on the underlying Chopper `Request` object. These can be read
in `Converter`s or `Interceptor`s for tracing, analytics, varying behavior, and more.


if want to filter null value or empty String for some url. we can make an `IncludeBodyNullOrEmptyTag` Object as Tag.
```dart
class IncludeBodyNullOrEmptyTag {
  bool includeNull = false;
  bool includeEmpty = false;

  IncludeBodyNullOrEmptyTag(this.includeNull, this.includeEmpty);
}

@get(path: '/include')
Future<Response> includeBodyNullOrEmptyTag(
    {@Tag()
    IncludeBodyNullOrEmptyTag tag = const IncludeBodyNullOrEmptyTag()});
```

get tag via `request.tag` in `Converter` or `Interceptor`:

```dart
class TagConverter extends JsonConverter {
  FutureOr<Request> convertRequest(Request request) {
    final tag = request.tag;
    if (tag is IncludeBodyNullOrEmptyTag) {
      if (request.body is Map) {
        final Map body = request.body as Map;
        final Map bodyCopy = {};
        for (final MapEntry entry in body.entries) {
          if (!tag.includeNull && entry.value == null) continue;
          if (!tag.includeEmpty && entry.value == "") continue;
          bodyCopy[entry.key] = entry.value;
        }
        request = request.copyWith(body: bodyCopy);
      }
    }
  }
}
```

JEuler:
LGTM

JEuler (approved):

techouse (approved):

PR #581, opened 2024-03-07
:bookmark: release v7.2.0

techouse:
# chopper

## 7.2.0

- #579

# chopper_generator

## 7.2.0

- #579

PR #589, opened 2024-03-23
:bookmark: release v7.3.0

techouse:
# chopper

## 7.3.0

- Add support for `@Tag` annotation ([#586](https://github.com/lejard-h/chopper/pull/586))

# chopper_generator

## 7.3.0

- Add support for `@Tag` annotation ([#586](https://github.com/lejard-h/chopper/pull/586))

### DIFF
--- a/chopper/pubspec.yaml
+++ b/chopper/pubspec.yaml
@@ -1,6 +1,6 @@
 name: chopper
 description: Chopper is an http client generator using source_gen, inspired by Retrofit
-version: 7.2.0
+version: 7.3.0
 documentation: https://hadrien-lejard.gitbook.io/chopper
 repository: https://github.com/lejard-h/chopper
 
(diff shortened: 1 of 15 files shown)


techouse:
The `chopper_generator` dry run will fail because I raised the required `chopper` version to `^7.3.0` which has not yet been published and can't therefore be accessed in the dry run.

The version had to be bumped because `chopper_generator` now relies on a new annotation.

Essentially it's the same "problem" as described in https://github.com/lejard-h/chopper/pull/581#issuecomment-1984534514

CC / @JEuler @Guldem @lwj1994

chopper/pubspec.yaml
```diff
@@ -25,7 +25,7 @@ dev_dependencies:
   lints: ">=2.1.1 <4.0.0"
   test: ^1.24.4
   transparent_image: ^2.0.1
-  chopper_generator: ^7.1.0
+  chopper_generator: ^7.2.0
```
Guldem, line 28: Not really necessary for the release I think. But shouldn't this also use the latest (unpublished) version `7.3.0`

techouse: It should but if I do that then the publishing mechanism won't work at all because `chopper_generator: 7.3.0` is not on pub.dev 😅

I'll bump it in a later release.

Guldem (approved):

JEuler (approved):

PR #594, opened 2024-04-04
:arrow_up: update Github actions

techouse:
This should fix all the CI warnings like
```
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: dart-lang/setup-dart@d6a63dab3335f427404425de0fbfed4686d93c4f. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Cache Pub hosted dependencies
-        uses: actions/cache@ab5e6d0c87105b4c9c2047343972218f562e4319
+        uses: actions/cache@v4
         with:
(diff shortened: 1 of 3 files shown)


techouse:
Looks like the it's still gonna complain about Codecov because **we can not update to v4** 🤷

```
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: codecov/codecov-action@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

Guldem:
That's annoying.. Would an alternative be a viable option? I'm not really familiar with code coverage services..

JEuler (approved):

PR #592, opened 2024-04-03
:zap: use qs_dart for query string encoding

techouse:
@diegotori noticed that a `Map<String, dynamic>` like this
```dart
{
  'filters': {
    r'$or': [
      {
        'date': {
          r'$eq': '2020-01-01',
        }
      },
      {
        'date': {
          r'$eq': '2020-01-02',
        }
      }
    ],
    'author': {
      'name': {
        r'$eq': 'Kai doe',
      },
    }
  }
}
```
did not get encoded properly to
```
filters[$or][][date][$eq]=2020-01-01&filters[$or][][date][$eq]=2020-01-02&filters[author][name][$eq]=Kai%20doe
```
but became this
```
filters[$or][]={date: {$eq: 2020-01-01}}&filters[$or][]={date: {$eq: 2020-01-02}}&filters[author][name][$eq]=Kai doe
```

This PR fixes this by using an external library [qs_dart](https://pub.dev/packages/qs_dart) which is a Dart port of [qs](https://www.npmjs.com/package/qs) which I've been painstakingly working on over the past couple of weeks. 🤓

This now means that the output of Chopper, while still backwards compliant, is now [qs](https://www.npmjs.com/package/qs) compatible.

The only **caveat** is `null` and empty `String` handling in query strings, which is a bit different in [qs](https://www.npmjs.com/package/qs) and therefore also [qs_dart](https://pub.dev/packages/qs_dart).

Please check the various test case changes to see these `null` and empty `String` differences.

---

Addresses #584
Supersedes #591

CC / @JEuler @Guldem @diegotori

PR #595, opened 2024-04-04
:bookmark: release v7.4.0

techouse:
:bookmark: release v7.4.0

# chopper

## 7.4.0

- #592

# chopper_generator

## 7.4.0

- #592

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Cache Pub hosted dependencies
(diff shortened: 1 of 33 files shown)


techouse:
The `chopper_generator` dry run will fail because I raised the required `chopper` version to `^7.4.0` which has not yet been published and can't therefore be accessed in the dry run.

The version had to be bumped because `chopper_generator` now relies on a new annotation.

Essentially it's the same "problem" as described in https://github.com/lejard-h/chopper/pull/581#issuecomment-1984534514

CC / @JEuler @Guldem @diegotori

Guldem (approved):

JEuler (approved):

PR #609, opened 2024-05-03
:bookmark: release v8.0.0

techouse:
# chopper

## 8.0.0

- #604
- **BREAKING CHANGE**:
  - #547
    - `RequestInterceptor` and Function `RequestInterceptor`s are removed 
    - `ResponseInterceptor` and Function `ResponseInterceptor`s are removed
    - See [Migrating to 8.0.0](https://docs.google.com/document/d/e/2PACX-1vQFoUDisnSJBzzXCMaf53ffUD1Bvpu-1GZ_stzfaaCa0Xd3WKIegbd1mmavEQcMT6r6v8z02UqloKuC/pub) for more information and examples
  - #610

# chopper_generator

## 8.0.0

- #604
- **BREAKING CHANGE**:
  - #547
    - `RequestInterceptor` and Function `RequestInterceptor`s are removed 
    - `ResponseInterceptor` and Function `ResponseInterceptor`s are removed
    - See [Migrating to 8.0.0](https://docs.google.com/document/d/e/2PACX-1vQFoUDisnSJBzzXCMaf53ffUD1Bvpu-1GZ_stzfaaCa0Xd3WKIegbd1mmavEQcMT6r6v8z02UqloKuC/pub) for more information and examples

## 3.0.0

- Require Chopper ^8.0.0

PR #610, opened 2024-05-07
:sparkles: add `onlyErrors` option to `HttpLoggingInterceptor`

techouse:
Add ability to `HttpLoggingInterceptor` to **only** log HTTP errors.

### Rationale
The `HttpLoggingInterceptor` amazing, however, due to its verbosity the level of output can sometimes seem overwhelming. Having the ability to only show verbose logs for HTTP errors is therefore in my mind a nice feature.

### Usage

Passing `onlyErrors: true` to the constructor will limit the logger to only log **responses** with an HTTP status code of 400 and above.

---
This PR is a response/solution to @Guldem's https://github.com/lejard-h/chopper/pull/609#issuecomment-2098284779

### DIFF
--- a/chopper/lib/src/interceptors/http_logging_interceptor.dart
+++ b/chopper/lib/src/interceptors/http_logging_interceptor.dart
@@ -3,6 +3,7 @@ import 'dart:async';
 import 'package:chopper/src/chain/chain.dart';
 import 'package:chopper/src/chopper_log_record.dart';
 import 'package:chopper/src/interceptors/interceptor.dart';
+import 'package:chopper/src/request.dart';
 import 'package:chopper/src/response.dart';
 import 'package:chopper/src/utils.dart';
 import 'package:http/http.dart' as http;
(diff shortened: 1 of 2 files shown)


JEuler (approved):

techouse:
@Guldem would you mind throwing an eye on this?

P.S. I've also replaced all the `String` concatenation and variable re-assignments with a `StringBuffer` for (hopefully) [better performance](https://medium.com/@SabinPoudel6969/string-vs-stringbuffer-optimizing-performance-in-dart-46d900b7d0de).

chopper/lib/src/interceptors/http_logging_interceptor.dart
```diff
@@ -75,116 +76,152 @@ enum Level {
...
+      );
     }
-    final stopWatch = Stopwatch()..start();
-
-    final response = await chain.proceed(request);
 
-    stopWatch.stop();
+    if (level == Level.none || (onlyErrors && response.statusCode < 400)) {
+      return response;
+    }
```
Guldem, line 165: Duplicate code? Its also at the beginning of the method

Guldem (changes requested):
Nice addition! Found 1 small thing

chopper/lib/src/interceptors/http_logging_interceptor.dart
techouse: Yeah, it was there in the original file https://github.com/lejard-h/chopper/blob/707950bbb3379ea2332b040815b9205c3184c40e/chopper/lib/src/interceptors/http_logging_interceptor.dart#L143 so I just added the additional condition. But on second thought, I should remove it.

Guldem (approved):

PR #634, opened 2024-08-29
:art: remove duplicate null-coalescing responseTypeReference

techouse:
While looking at the generator I found a duplicate null-coalescing `responseTypeReference`. Probably left there by mistake. It's not a bug per-se, it just looks up the response type 2x for no reason.

CC/ @Guldem


JEuler (approved):

Guldem (approved):
Small woepsie :)

PR #645, opened 2024-12-16
add option to use cancellation token in ChopperClient

kfiross:


techouse:
Hey,

I love the idea of a cancellation token, however, I'm not a fan of using a fork of the http package by default.

Could you maybe create an opt-in plugin instead?

PR #661, opened 2025-05-22
add some changes in [ChopperException] class 

ahmedtalal:
Changed in ChopperException class to handle HTTP errors in a clear and structured way. This improves error reporting and makes it easier to manage failed responses gracefully.
Kindly consider accepting the request if it fits the direction of the project. 🙏

### DIFF
--- a/chopper/lib/src/chain/interceptor_chain.dart
+++ b/chopper/lib/src/chain/interceptor_chain.dart
@@ -48,7 +48,12 @@ class InterceptorChain<BodyType> implements Chain<BodyType> {
     if (index + 1 < interceptors.length &&
         interceptor is! InternalInterceptor) {
       if (response == null) {
-        throw ChopperException('Response is null', request: request);
+        throw ChopperException(
+          request: request,
+          type: ChopperExceptionType.unknown,
+          error:
+              'Interceptor [${interceptor.runtimeType}] did not return a response',
+        );
       }
 
       assert(
--- a/chopper/lib/src/chopper_exception.dart
+++ b/chopper/lib/src/chopper_exception.dart
@@ -1,24 +1,186 @@
 import 'package:chopper/src/request.dart';
 import 'package:chopper/src/response.dart';
 
-/// {@template ChopperException}
-/// An exception thrown when something goes wrong with Chopper.
-/// {@endtemplate}
+/// Represents the type of error occurred in Chopper.
+enum ChopperExceptionType {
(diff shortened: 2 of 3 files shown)


techouse:
Hi,

While I appreciate the effort, this is far too naive.

There are [a couple dozen of HTTP exceptions](https://api.dart.dev/stable/2.4.1/dart-io/HttpStatus-class.html) to handle and a simple enum just won't do

You'd need to take care of all of these by creating specific HttpException classes.

chopper/lib/src/chopper_exception.dart
```diff
@@ -1,24 +1,186 @@
...
+  connectionTimeout,
+  sendTimeout,
+  receiveTimeout,
+  badResponse,
+  cancel,
+  connectionError,
+  unknown,
+}
+
+extension ChopperExceptionTypeDescription on ChopperExceptionType {
```
techouse, line 15: With enhanced enums in Dart 3, creating extensions like these is no longer necessary. You can simply add a the method `toPrettyDescription` to the `ChopperExceptionType` enum.

chopper/lib/src/chain/interceptor_chain.dart
```diff
@@ -48,7 +48,12 @@ class InterceptorChain<BodyType> implements Chain<BodyType> {
     if (index + 1 < interceptors.length &&
         interceptor is! InternalInterceptor) {
       if (response == null) {
-        throw ChopperException('Response is null', request: request);
+        throw ChopperException(
+          request: request,
+          type: ChopperExceptionType.unknown,
+          error:
```
techouse, line 54: What's the benefit of all of this? The type is `unknown` and the error is basically the same.

chopper/lib/src/interceptors/http_call_interceptor.dart
```diff
@@ -30,7 +30,11 @@ class HttpCallInterceptor implements InternalInterceptor {
       final response = await http.Response.fromStream(streamRes);
       return Response(response, response.body as BodyType);
     } else {
-      throw ChopperException('Unsupported type', request: chain.request);
+      throw ChopperException(
+        request: chain.request,
+        type: ChopperExceptionType.unknown,
+        error: 'Unsupported response type: ${BodyType.runtimeType}',
```
techouse, line 36: Rather than writing a "pretty" message, this the `ChopperException` should consume a `StackTrace`

